PR #4, opened 2018-05-31
Update taskflow.cpp

NanXiao:
Fix build error:  
```
error: variable 'num_tasks' cannot be implicitly captured in a
      lambda with no capture-default specified
```


clin99:
Could you provide the compiler version that gives the build error?

Thanks.

NanXiao:
@clin99  

ArchLinux, [gcc-8.1.0](https://www.archlinux.org/packages/core/x86_64/gcc/). Thanks!

PR #7, opened 2018-06-01
Update CMakeLists.txt

NanXiao:
Maybe should not hardcode compiler in `CMakeList.txt`, otherwise it seems no method to select `clang`. Sorry if I am wrong, thanks!


twhuang-uiuc:
Thanks for the catch-up! Currently we use g++ because it's simple to just use `fopenmp` flag to compile examples and unittests. However, as far as I know it requires another library (`clang-omp`) to use OpenMP in clang. Correct me if I am wrong.

NanXiao:
If you don't hardcode, the CMake will select the default compilers in your OS. On Linux, I think it should be `gcc/g++` in most distros. So if user want to switch to other compilers, he can:  

     cmake -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++  ..

But if you hardcode it, the user can't switch to `clang`, and the only method seems to modify your `CMakeList.txt`. Thanks!

twhuang-uiuc:
We have merged and updated a new `CMakeList.txt` to reflect this problem. Thank you!

PR #33, opened 2018-08-27
Add threadpool.hpp (C++14 "light" version, only including tf::Threadpool)

totalgee:
- if you're using a C++17 compiler, you should include the main
  "taskflow.hpp" header. However, if you're working on a C++14 project
  and only require the Threadpool functionality, you may include
  "threadpool.hpp".


totalgee:
This is just a first PR to see if this is something you'd be interested in having in the library...for me, it's very useful (as discussed offline with @twhuang-uiuc) to have a C++14 version, even one that only includes limited functionality -- in particular, the `tf::Threadpool` class.

I don't think you should necessarily "mess up" the cleanness of the main library by trying to make it backwards compatible with pre-C++17 versions...the clean interface and code is part of what attracted me to the project!  However, for my purposes, it would be handy to have the Threadpool available to projects that are still C++14-dependent...hence this code offering.  If it has no place in the main project, that's fine, I can just use it myself.

totalgee:
If it is to go into the project, I'd have to add some unit tests and an example.

Issue #21, opened 2018-06-16
Visual Studio/Windows support


vblanco20-1:
Hello. I went to try this library on windows, but ive had some issues with visual studio 2017 latest update. 

With a quick source code edit, i was able to have the matrix example running well, but thats a hack.
When compiled on visual studio, the library throws a huge amount of warnings.  They are mostly about type casting and data padding.

![image](https://user-images.githubusercontent.com/1417000/41497405-07fce682-7155-11e8-99e4-094e0a3b8070.png)

The library fails to compile due to Threadpool::async<>, wich somehow fails the second if constexpr
(Line 279 of taskflow.hpp). The branch is ignored and it tries to call .set_value(c) when the std::promise is of void type. By commenting that out, the library compiles and i can run the matrix demo with full parallism. The issue is that doing that is a huge hack.

twhuang-uiuc:
It seems to be the problem of `if constexpr`. I did a quick check and found [here](https://blogs.msdn.microsoft.com/vcblog/2017/12/19/c17-progress-in-vs-2017-15-5-and-15-6/) you need at least VS 2017 15.3 to support this compile-time switch. Would you please verify your version? 

As of yet, we havn't checked on MS platform but Linux and MAC OS. We will support it soon.

vblanco20-1:
@twhuang-uiuc  I have the absolute latest version of VS. it does have if constexrp, and it even works on the *other* if in the function. For some reason is the inner if the one that doesnt work well, not the outer if. Maybe a compiler issue? 

I fixed it myself by moving that constexrp if into a templated function, it seems to work, and i can run the tests with the fixes.

I can do a pull request if you want with the edit.

Another thing i was thinking about. Can you schedule tasks inside tasks? Like, for example, i have a task that might or might not perform an internal parallel for, depending on some internal task data. Can i schedule some "child" tasks inside that task to do the parallel for?

twhuang-uiuc:
Interesting, would you please give a pull request so I can take a look at this fixes? At the same time, I will go get a MS machine to test it out. Hopefully this can be solved next week.

I absolutely agree with your idea of scheduling tasks inside tasks. We are still thinking about an interface to incorporate this intra-task parallelism. This can be very useful when I need to spawn an inner parallelism during an task execution context. Feel free to give me any of your thoughts regarding the API, spec you would like to see.

Thanks for your insightful feedback.

vblanco20-1:
The pull request is on #22 .

[...]

PR #35, opened 2018-08-28
Add threadpool_cxx14 example; fix all Windows build warnings

totalgee:
With these changes, the Windows (VS 2017) version compiles without any warnings using CMake, and all tests run successfully as well.

  - use %zu format rather than %lu for printf with size_t types.
  - explicitly throw away "nodiscard" return value of std::adjacent_find().
  - explicitly cast to unsigned type when creating a Taskflow.
  - (git) ignore Windows/MSVC files and directories.

I also added a simple example showing the usage of `threadpool_cxx14.hpp`.


totalgee:
I think this should allow closing issue #21.

totalgee:
I added another commit, including a unit test for the threadpool_cxx14.hpp file.

Issue #36, opened 2018-08-30
Can this be used to run a "single" task a thousand times? (each time with different arguments)


markg85:
Hi,

I have a fairly simple task: make thumbnails of a thousand large images.
Right now i'm using a thread pool which works great! I basically have one function that accepts two input arguments (the source of the large image and the destination of the thumbnail), that is then added to a queue which is then emptied by the threadpool.

But how would you do such a thing in this tasks library? Is it even intended to be used like that, and if so, how?
An example would be really great :)

I've been looking at the parallel_for example but i don't quite see how i would apply that to my case yet.

Cheers,
Mark

twhuang-uiuc:
Hi, if your application has no dependency and you just need a pool of threads to do your job simultaneously, you can just use our built-in threadpool.

```cpp
tf::Threadpool tp(4);  // create a threadpool of 4 worker threads

tp.silent_emplace([] () { /* task 1*/ });    // silent_emplace won't return a future
tp.silent_emplace([] () { /* task 2*/ });
... add more tasks

std::future<int> fu1 = tp.emplace([](){ return 1; })  // emplace returns a future
... add more tasks
```

totalgee:
Just a clarification about why you added the `future<int>` at the end of your threadpool example...I assume you did this so you could "check" when all the tasks have completed, without creating futures for each of them?  (As an unrelated comment, it wouldn't have to be `future<int>`; it might as well just be `future<void>`.)

However, calling "get" on this task's future won't guarantee that all the earlier-scheduled tasks have completed, right? Tasks are scheduled in the pool of workers in order they're "emplaced", so that last-added task will indeed be *scheduled* last to a worker thread, but there's no guarantee the last few other tasks will have finished by the time its future is ready...

Maybe what you mean is: since `tp.shutdown()` is called implicitly by the `tf::Threadpool` destructor, as long as all tasks in the queue have been assigned to a worker already(!), this will ensure all tasks are completed when the shutdown method finishes running.

So you'd have to add to your example (if it's important to ensure all the tasks are complete at a certain point):
```cpp
  fu1.get();
  // At this point it's still not guaranteed all tasks have completed...
  tp.shutdown();  // or just leave tp's scope so its destructor is called!
  // ...at this point, all the tasks should be done.
```

Do I understand correctly?

twhuang-uiuc:
[...]

PR #37, opened 2018-08-31
Add wait_for_all() method to Threadpool

totalgee:
(and add a unit test for this feature)

This was suggested (by me ;-) in issue #36.


totalgee:
I'm not sure if this is necessary or desired … cost-wise, probably it's about the same to actually destroy the old Threadpool instance and create a new one whenever you want to "flush" all scheduled tasks, since this implementation does it by shutting down and respawning a new set of worker threads.  But conceptually I think it's nice to have the same `wait_for_all` method that Taskflow does (for simple use cases).

It could be implemented in a more optimal way (without shutting down the threads -- perhaps by adding new signalling, or some way to tell them to temporarily stop pulling new tasks from the queue), but with that would add more complexity to the (currently very easy-to-follow) Threadpool design.

What do you think?

twhuang-uiuc:
Hi @totalgee  I think there should be a better solution to do this without shutting down the threadpool?

totalgee:
Okay, I just uploaded a better version that doesn't shut down/restart worker threads.  It uses barrier-like synchronization to ensure they've all completed the last set of tasks before moving on.

PR #38, opened 2018-09-03
ProactiveThreadpool

gguo4:
I have added ProactiveThreadpool and its associated test examples

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -96,6 +96,9 @@ target_link_libraries(emplace Threads::Threads)
 add_executable(reduce example/reduce.cpp)
 target_link_libraries(reduce Threads::Threads)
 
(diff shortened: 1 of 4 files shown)


totalgee:
Interesting. Is there a reason for a new TP rather than changing the behaviour of the existing one?

Also, could you add the wait_for_all() method (see PR #37), to keep the interface the same for all thread pools?  Thanks!

gguo4:
Yes, we are experimenting on how a different TP scheduler may affect the throughput of job completion. The new TP will be twice faster than the old one if we are inserting lightweight jobs like empty jobs and atomic adds. This can be tested by the new threadpool example. 
Yes, I am examining the wait_for_all method. It will be added to the new TP.

PR #68, opened 2019-02-25
Add an example about a way to pass data between tasks

paolobolzoni:
Here is an example how to pass data from tasks to their successors and to use cpp-taskflow for synchronization.

In real world code probably some kind of factory class will ensure all the invariants, but I think it is quite small to show the point.

### DIFF
--- a/example/data_flow.cpp
+++ b/example/data_flow.cpp
@@ -1,0 +1,154 @@
+/*
+
+cpp-taskflow works on directed acyclic graphs.
+And here we want to pass information between the flow elements.
+
+To do so, we see the cpp-taskflow arcs as objects memory where the functions on
+the nodes read from or write to.
+
+The function on every node will *read from* the objects of memory of its
+incoming edges and *write to* the objects of its outcoming edges.
+
+The cpp-taskflow semantics ensures the synchronization.
+
+
+Nodes without incoming edges will require the input from somewhere else;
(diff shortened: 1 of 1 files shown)


twhuang-uiuc:
Hi @paolobolzoni , I have merged your example to both dev and master branches. I also highlighted your contributions in the README. Thank you!

In the future, we can probably come up with a more general dataflow interface/API to taskflow.

paolobolzoni:
I would be great to have a direct way to express dataflows directly.

Still I am afraid it would be really difficult, the problem I see is that in cpptaskflow to define the functions for the nodes (eg, `tf.emplace`) and set up the arcs (eg, `.precede`) is done in two steps. This is the normal way to create graphs, it is convenient, and easy to understand.

But it makes more difficult to have a general way to express how many objects are "input" and "output" in the nodes functions because when you define the function you still do not know how many incoming and outcoming arcs you will have.

As you can see in the example, I basically set up the objects for the arcs (the three vectors), define the taskflow functors (via classes with `operator()`), and finally create the taskflow arcs (via `.precede`).
Ideally one wants to have compile time error if the taskflow graph does not match the functions; for example the function expects two incoming edges and there is only one. But I cannot see a general way to do so.

Of course I will be happy to be proven wrong. After all, cpptaskflow as-is already is quite the accomplishment.

PR #73, opened 2019-03-27
Add a ready-to-use generic class to associate data to the flows

paolobolzoni:
The class is a thin wrapper around taskflow that ensures that each arc in the taskflow DAG has a associated object (of type T) and allows nodes to execute a function that can easily read from the incoming arcs objects and write in the outcoming arcs objects.

All arcs are associated with an object of the same type T. Since the graph is built runtime it would be really difficult to do otherwise. If one needs different types of objects on the arcs, then one can use class hierarchies and virtual functions and the type T would be a unique_ptr<some base class>. Otherwise one can use std::variant.

The new example is an extension of the previous one. It wants to show that with a bit of boiler plate code that give human-readable names to the objects on the incoming and outcoming arcs it is possible to have readable parallel functions, with input and output, synchronized by the taskflow semantics.

I also renamed my previous example and fix the English language a bit.

### DIFF
--- a/example/dataflow.hpp
+++ b/example/dataflow.hpp
@@ -1,0 +1,228 @@
+#ifndef DATAFLOW_TPP_HPP
+#define DATAFLOW_TPP_HPP 1
+
+#include <taskflow/taskflow.hpp>
+
+#include <utility>
+#include <vector>
+#include <functional>
+
+namespace df {
+
+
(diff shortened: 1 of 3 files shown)


tsung-wei-huang:
Hi @paolobolzoni , interesting idea! I will check the file again and merge it to our branch if no problem.

PR #80, opened 2019-04-23
Fix unused-lambda-capture clang-warning

soonho-tri:
To fix the following warning (from clang 6):

```
cpp-taskflow/taskflow/executor/workstealing.hpp:492:37: warning: lambda capture 'N' is not used [-Wunused-lambda-capture]
    _threads.emplace_back([this, i, N] () -> void {
                                    ^
```


tsung-wei-huang:
Thanks!

PR #83, opened 2019-04-30
gitignore, observer fix, observer names

pgruenbacher:
i wouldn't necessarily merge this... but I wanted to illustrate some things...

the executor observer would really benefit from being able to output the names of the tasks if the tasks have a node associated with them. Makes the tracing infinitely more useful.

idk what text editor you use but there's lots of trailing spaces in the files I had edited.

there's no gitignore which is problematic when working within build directory or the executables being built in the examples dir...


pgruenbacher:
oh and there's a fix in there in the observer dump method for when there's multiple empty worker threads causing there to be no comma between brackets.

tsung-wei-huang:
Hi @pgruenbacher , thanks for your contributions. I don't quite think it's a good idea to put a "std::to_string" in this very bottommost call. I prefer to leave it to users who call always derive their own observer class from `tf::ExecutorObserverInterface`.

pgruenbacher:
@tsung-wei-huang I agree, the issue is that the interface to observer is just `on_entry(worker_id)`
when it would instead need to be something like `on_entry(worker_id, node)`. I have a hard time believing anyone is using observer as is currently because it's not useful without having the name of the executing nodes.

tsung-wei-huang:
Hi @pgruenbacher , I have modified the observer to take additional argument on `TaskView`. TakView is a lightweight immutable wrapper to the internal node but allows you to query the attributes of the task node such as it's name and dependency count. Would you please take a loot at [observer.hpp](https://github.com/cpp-taskflow/cpp-taskflow/blob/master/taskflow/core/observer.hpp)?

pgruenbacher:
yes that works for me, now I'll be able to subclass the observer to match what I expect. thanks.

tsung-wei-huang:
> yes that works for me, now I'll be able to subclass the observer to match what I expect. thanks.

Thanks for your support.

PR #101, opened 2019-07-29
Fix potential missing comma when dumping thread activity

mrogez-yseop:


tsung-wei-huang:
Thanks @mrogez-yseop !

PR #109, opened 2019-08-09
Fix cmake Apple edge case

mdaiter:
(Prereq is using CMake 3.0.0 or higher)
Using MATCHES instead of STREQUAL fixes the Clang comparison for checking compiler support in the CMake script.
https://stackoverflow.com/questions/10046114/in-cmake-how-can-i-test-if-the-compiler-is-clang


mdaiter:
Not sure why continuous integration's failing for the Microsoft build...seems like a you problem, not a me problem ;)

PR #110, opened 2019-08-09
Fixed build for MacOS 10.13 and prior

mdaiter:
Apple 10.13 and prior disable `std::get` for multiple reasons. This patch creates a workaround for compilations on MacOS 10.13 and prior. Not sure how you want to handle null pointers returned from `std::get_if` or if you want to force users to use MacOS 10.14 and later.


mdaiter:
(For a more comprehensive review: https://stackoverflow.com/questions/52521388/stdvariantget-does-not-compile-with-apple-llvm-10-0?rq=1 )

PR #112, opened 2019-08-12
Removed redundant constexpr-if

AlexKiryushkin:
Removed redundant constexpr-if when advancing the iterator since it is done by std::advance implementation

### DIFF
--- a/taskflow/core/flow_builder.hpp
+++ b/taskflow/core/flow_builder.hpp
@@ -338,16 +338,8 @@ std::pair<Task, Task> FlowBuilder::parallel_for(I beg, I end, C&& c, size_t p){
 
     auto e = beg;
     size_t g = (w++ >= r) ? b - 1 : b;
-    
-    // Case 1: random access iterator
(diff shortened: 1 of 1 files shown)


jonesmz:
> Hi @AlexKiryushkin , for non-random access iterator, both `std::distance` and `std::advance` take linear time complexity. So we use a `if_constexpr` to separate the advancement of different iterators for efficiency.

Could you go into more detail about how 

```cpp
for(size_t i=0; i<g && e != end; ++e, ++i);
```

Is any different than `std::distance` and `std::advance` ? They for loop is also linear time complexity, is it not?

However, I do notice that you're calculating `std::distance(beg, end) in a loop in at least one of the functions in this PR.  That seems like an unnecessary calculation, since you can use the distance calculated at the beginning and subtract the number of items you've traversed so far. Saving on recalculating the distance every loop.

AlexKiryushkin:
Hello.

I guess, this PR is outdated. So, probably, it might be closed.

PR #137, opened 2020-01-16
size_t instead of int

matt77hias:
This eliminates quite some warnings for MSVC++.


dcnieho:
I would like to see this change as well, getting quite some warnings

tsung-wei-huang:
@matt77hias @dcnieho  thanks for the pull requests. It took a while to merge because the previous scheduler relied on a negative trick to minimize the counter overhead of the scheduler. Now it's clear size_t is the best. I have merged the pull request. Thanks

PR #146, opened 2020-02-28
Fix typos and paths

leokolln:
- Point to correct folder `examples` instead of `example`.
- Fix code snippet with call to `run_until(...)`.


tsung-wei-huang:
@leokolln  Thanks for the catch!!

PR #162, opened 2020-04-27
strcmp needs to check for 0 for equality.

iplayfast:
was checking for inequality before.


tsung-wei-huang:
thanks for the catch!

PR #167, opened 2020-05-14
size_t instead of unsigned

matt77hias:
Related to https://github.com/cpp-taskflow/cpp-taskflow/issues/166.


tsung-wei-huang:
@matt77hias Thanks!

Issue #165, opened 2020-05-07
Possibility to run "Fire-and-forget" Taskflows


robinchrist:
**Please describe your feature request.**
I'd like to enqueue Taskflows to an executor in a Fire-and-forget fashion, i.e. not having to care about their lifetime manually, but rather have the library delete them once they have finished execution.

Is it possible to implement that? Or am I thinking way too complicated? 

At what **exact** point in time can the Taskflow be destroyed without risking nullptr dereference by the lib or something similar?

robinchrist:
I basically need a callback with a strong "All cleanup tasks are done, nothing refers to the taskflow anymore" guarantee.
As far as I can tell from the source code, it's not safe to delete the Taskflow in the callback function you can pass to the executor:
https://github.com/cpp-taskflow/cpp-taskflow/blob/75500eccbca3049c06a63ad5aa57e27db348e2d5/taskflow/core/executor.hpp#L1112

To me it seems that the callback is called but after the call has completed, references to the Taskflow are still made. Right?

tsung-wei-huang:
Hi @robinchrist , I believe this is doable with some hacks. Topology is a data structure created by the executor at runtime when you run a taskflow. It does store the callback and that callback lives together with the topology. You can store your taskflow in the callback capture, something like the following:

```cpp
auto taskflow = std::make_shared<Taskflow>();
executor.run(*taskflow, [taskflow] () { std::cout << "end of the execution and taskflow is still safe\n";  }).wait();
```

robinchrist:
Thanks for the info!
Just to be sure and sum things up: Once the callback has been called and exited, you can safely delete the Taskflow without risking access to already freed memory, right?

tsung-wei-huang:
Yes. You can take a look at the [code](https://github.com/cpp-taskflow/cpp-taskflow/blob/7cc2f3872df939f3f9abbdf237474f2b2c6d208d/taskflow/core/executor.hpp#L1185). When the callback is called, executor check if it is the last topology (due to `run_n`) and then erase the topology from the list [here](https://github.com/cpp-taskflow/cpp-taskflow/blob/7cc2f3872df939f3f9abbdf237474f2b2c6d208d/taskflow/core/executor.hpp#L1224). 

Hope this help clarify your question :)

fstrugar:
[...]

PR #189, opened 2020-06-19
Fix for https://github.com/taskflow/taskflow/issues/165#issuecomment-…

fstrugar:
…645690025 (mutex unlock gets called after the topology object gets destroyed here).


tsung-wei-huang:
@fstrugar thanks for your pull request!

PR #196, opened 2020-07-14
Compile with /bigobj on MSVC

remz1337:
This PR fix an issue with x64-windows. Compile with /bigobj for x64-windows target. I will also use this as a patch to update the VCPKG port


tsung-wei-huang:
Perfect. Thanks!

Issue #191, opened 2020-06-24
Allow user control of thread creation


xbreak:
**Please describe your feature request.**

I'd like to create my own _host_ threads used by `tf::Executor`, that can be created using the full capability of my platform, to set different policies (NUMA affinity et.c.). In one of my use-cases I would then have differently "flavored" `Executors` instances, each with different properties. In contrast to #190 I'm not asking for any scheduling support (if I understood that request correctly). I would be responsible for submitting tasks to the appropriate executor which I think makes this feature low-impact.

To make it more concrete one use-case I have is to isolate a `tf::Taskflow` computation to a given NUMA node to avoid interference with other activities on other nodes. To do that I'd create a set of threads that match the NUMA topology, set memory policy, affinities and so on.

Given the `std::thread` API a feasible option would be if `tf::Executor` accepts a callable that creates the threads, maybe something like (pseudo-code-ish):

```c++
template<class ThreadCreator>
explicit Executor(ThreadCreator tc, size_t N = std::thread::hardware_concurrency());
...
_spawn() {
  ...
  _threads.emplace_back(
      std::invoke(tc,
                  [this] (Worker& w) -> void {...}
      )
  );
}
```
where `ThreadCreator` implements the concept matching `std::thread` constructor, so something like:

```c++
template< class Function, class... Args >
std::thread (Function&& f, Args&&... args);
```

As far as I can tell the threads are only spawned once so there would be no need to store `ThreadCreator` so no need to make `tf::Executor` class-templated or resort to type erasure. As far as I can tell only `_spawn` would be affected.

Cheers, and thanks for a very nice library!

tsung-wei-huang:
@xbreak thanks for your questions. If I understand correctly, you are asking to have executor-level thread binding. Your idea of passing thread creator is indeed flexible but I am not sure how it is used in the context of template in executor? Probably it's better to let users define upon constructor which thread binds to which cpu?

xbreak:
@tsung-wei-huang: Thanks for the quick reply! 

> If I understand correctly, you are asking to have executor-level thread binding.
Yes, and more.

> ... but I am not sure how it is used in the context of template in executor?

Would you mind elaborating on this? I'm not quite following.

>  Probably it's better to let users define upon constructor which thread binds to which cpu?

Do you mean passing in a data structure that describes e.g. affinities? 
[...]

Issue #160, opened 2020-04-20
Limiting Concurrency


jlaxson:
Hi - I really like what you've built here.  I have a use case that I can't quite figure out how to map to cpp-taskflow, hoping I could get some advice.

In my workflow I have a number of steps that I'd like to artificially limit concurrency - for instance, a subflow that consumes a very large amount of memory.  Or that consume limited external resources (gpus, etc).  Or even to differentiate between CPU-bound tasks (at most one worker per cpu thread) and IO bound tasks (perhaps many more than one per thread?).

I think I could emulate this with mutexes or other synchronization outside of taskflow itself, but that has the side effect of consuming/blocking workers when they could be working on other ready tasks.

One potential solution would be to have the ability to assign tasks/subflows to specific executors, with dependencies across the executors.  So I could have my GPU executor, with concurrency 2, in parallel to the CPU executor with native hardware concurrency.  Syntax could look something like:

```c++
tf::Executor cpu;
tf::Executor heavyResourceExecutor(1);
tf::Taskflow tf;

for (int i = 0; i < 100; i++) {
  auto load_task = tf.emplace([](){ /* load some data */ });
  auto heavy_task = tf.emplace(heavyResourceExecutor, []() { /* do some heavy lifting */ });

  load_task.precede(heavy_task);
}

cpu.run(tf); // load tasks can happen at full concurrency, but heavy_task will happen one at a time
heavyResourceExecutor.run(); // starts its worker threads, but won't do anything until a heavy_task is ready
```

Wondering if there's a way to achieve this with the functionality you have now?  Otherwise, a humble feature suggestion.

tsung-wei-huang:
Hi @jlaxson , I think in your case you can just use one `cpu` executor to do both `load_task` and `heavy_task`. Our runtime scheduler is adaptive; it won't consume all the threads all the same. If you really want only one thread to run `heavy_task`, you can artificially create a set of dependency to force `heavy_task` to become a super dependent/successor of other tasks.

jlaxson:
Hi, thanks for your quick reply.

>  Our runtime scheduler is adaptive; it won't consume all the threads all the same but adapts the number of threads to available task parallelism at any time during the graph execution. 

I had looked a bit through the code but all I see is Executor::_spawn which is always used to spawn a constant number of threads.  Can you point me towards more detail?

> If you really want only one thread to run heavy_task, you can artificially create a set of dependency to force heavy_task to become a super dependent/successor of other tasks.

[...]

PR #206, opened 2020-08-07
Support custom domain

jlaxson:
This is a simple way to address my desires in #160.  Maybe also starts to speak to #191.  Not sure it's exactly what you'd want to add to the api but wanted to share a possible way forward.  I think my ideal case would be to be able to separate this from executors and be able to have the domains (or some concurrency object) dynamically defined by client code.

By the way, I had lots of issues with trailing whitespace preparing this diff.  Would you be interested in a PR that cleans that up across the repo?

### DIFF
--- a/taskflow/core/executor.hpp
+++ b/taskflow/core/executor.hpp
@@ -59,7 +59,7 @@ class Executor {
     /**
     @brief constructs the executor with N worker threads
     */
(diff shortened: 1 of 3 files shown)


tsung-wei-huang:
Hi @jlaxson thanks for the suggestion. What do you mean bu trailing space? I will be happy to see a pull request that removes any issues of that.

PR #208, opened 2020-08-13
Added static_cast for double to size_t conversion

matt77hias:


tsung-wei-huang:
@matt77hias thanks for the catch!

PR #210, opened 2020-08-13
C4458

matt77hias:


matt77hias:
I just replaced the `i` (of increment) with the `a` (of addend), as it is clearer to preserve the `i` in the for-loops.

tsung-wei-huang:
@matt77hias thanks!

Issue #229, opened 2020-09-22
Wrong test case called (uuid instead of uuid.10threads)


musteresel:
https://github.com/taskflow/taskflow/blob/693ab005203fe1fccc1ffb974b2ffad85a60f1bc/CMakeLists.txt#L370

Issue #226, opened 2020-09-17
Allow out-of-tree builds


musteresel:
**Allow (what is normally standard with CMake) out-of-tree builds**

I was *very* surprised when I first compiled taskflow with my usual approach to a CMake based project:

~~~
# in taskflow repo
cmake -S . -B builddir
cmake --build builddir
cd builddir
ls
~~~

You deliberately break the usual CMake flow by modifying the output directory to put the example/test executables in the source tree. **Why?**

Supporting out-of-source/out-of-tree builds has many positive side effects:

 - Source tree can be read only (some packaging systems require this; or dedicated build servers accessing a network share which is read only for security reasons)
 - Multiple builds (different options / compilers) can happen in parallel; can be tested in parallel
 - Source tree remains clean of build files
 - Removing build files means just removing the build directory

Would you accept a PR which fixes this?

tsung-wei-huang:
@musteresel MANY THANKS if you can submit a PR to fix it! This is one of my TODO items but somehow it was left blank.

PR #230, opened 2020-09-22
General build system improvements

musteresel:
- Out-of-source builds:  This commit reduces the manual adjustment of
   the CMake output path and instead uses the subdirectories with
   their own CMakeLists.txt files to let CMake handle build
   directories automatically  (https://github.com/taskflow/taskflow/issues/226)

 - Many executables are specified to be built in exact the same way;
   this commit uses foreach loops in most of these cases to reduce
   overall amount of CMake code

 - doctest test discovery: doctest ships a CMake utility to
   automatically discover and add any tests within a doctest
   executable.  This is highly perferrable to manually specifying all
   the individual tests (which is error-prone and adds visual clutter)

 - Fixes https://github.com/taskflow/taskflow/issues/229 (wrong test
   case called: uuid instead of uuid.10threads)

Todo:

 - [ ] TBB currently still "pollutes" the source directory with CMake files; I'm not sure why / how to approach this
 - [ ] Better documentation of the build process
 - [ ] Further reduce repetition when executables are built in a similar way

This is a quite large commit; I wasn't sure whether breaking it down into smaller commits would bring that much benefit, though?


tsung-wei-huang:
@musteresel thanks for your effort - yes, tbb is a particular issue there. Taskflow included the tbb source in the 3rd-party directory because we were doing research experiments and we need to stick with a particular version (and manuscript submission). That being said, for now, I would prefer to remove it and ask users to install tbb (with a version requirement) on their machines if they want to compile benchmarks. This will resolve the pollution issue and make the entire project directory light.

I will merge this PR later today.

PR #234, opened 2020-10-02
Added detection of Solaris (and illumos) in utility/os.hpp

vedranmiletic:


tsung-wei-huang:
@vedranmiletic  thanks!

Issue #242, opened 2020-11-03
Taskflow library build fails when included after <windows.h> header


Croolman:
**Describe the bug**
In a simple project, when `windows.h` header is included in project before the `taskflow/taskflow.hpp` header, the build fails.

**To Reproduce**
Create project 
* include `windows.h` header
* include `taskflow/taskflow.hpp` header after that
* try to build

**Desktop (please complete the following information):**
 - OS: Windows 10

**Additional context**
Is this a bug or a documented limitation?

tsung-wei-huang:
@Croolman interesting, do you have the compile error message? I may not have access to windows shortly.

Croolman:
@tsung-wei-huang sure, here it is:

```
taskflow\utility\object_pool.hpp(68,36): error C2589: '(': illegal token on right side of '::'
taskflow\utility\object_pool.hpp(198): message : see reference to class template instantiation 'tf::ObjectPool<T,S>' being compiled
taskflow\utility\object_pool.hpp(68,36): error C2059: syntax error: '('
taskflow\utility\passive_vector.hpp(176,73): warning C4003: not enough arguments for function-like macro invocation 'max'
taskflow\utility\passive_vector.hpp(176,73): error C2589: '(': illegal token on right side of '::'
taskflow\utility\passive_vector.hpp(209): message : see reference to class template instantiation 'tf::PassiveVector<T,S,A>' being compiled
taskflow\utility\uuid.hpp(78,41): warning C4003: not enough arguments for function-like macro invocation 'min'
taskflow\utility\uuid.hpp(78,41): error C2589: '(': illegal token on right side of '::'
taskflow\utility\uuid.hpp(78,41): error C2059: syntax error: ')'
taskflow\utility\uuid.hpp(78,41): error C2143: syntax error: missing ',' before '?'
taskflow\utility\uuid.hpp(78,41): error C4430: missing type specifier - int assumed. Note: C++ does not support default-int
taskflow\utility\uuid.hpp(78,41): error C2059: syntax error: '?'
taskflow\utility\uuid.hpp(79,41): warning C4003: not enough arguments for function-like macro invocation 'max'
taskflow\utility\uuid.hpp(79,41): error C2589: '(': illegal token on right side of '::'
taskflow\utility\uuid.hpp(80,3): error C2059: syntax error: ')'
taskflow\utility\uuid.hpp(83,42): error C2660: 'distribution': function does not take 1 arguments
[...]

PR #243, opened 2020-11-04
Prevent min max macros expansion when windows.h header is included in project alongside taskflow

Croolman:
Fix for the reported issue #242


tsung-wei-huang:
@Croolman thank you!

PR #246, opened 2020-11-16
Update executor.hpp

MambaWong:


tsung-wei-huang:
Thanks for the catch.

PR #249, opened 2020-11-22
Amend "settting" typo in docs

largerock:
Setting is spelt with 3 t's in some docs.
Replaces all instances of "settting" with "setting"


tsung-wei-huang:
@largerock thanks!

Issue #248, opened 2020-11-21
Support for intel compilers


trevor-vincent:
Hi, thank you for the really cool library! The CmakeLists.txt blocks intel compilers (icpc), e.g.

```
else()
  message(FATAL_ERROR "\n\
Taskflow currently supports the following compilers:\n\
  - g++ v7.0 or above\n\
  - clang++ v6.0 or above\n\
  - MSVC++ v19.14 or above\n\
  - AppleClang v8 or above\n\
")
```

It would be nice to have icpc support if possible.
Thanks!

tsung-wei-huang:
@trevor-vincent thanks for bringing this up! I think it's supportable by just adding an extra cmake else. Taskflow is standalone relying on only standard libraries. Would you be interested in submitting a pull request?

trevor-vincent:
> @trevor-vincent thanks for bringing this up! I think it's supportable by just adding an extra cmake else. Taskflow is standalone relying on only standard libraries. Would you be interested in submitting a pull request?

Excellent, I'm definitely interested, I'll get a pull request done in the next day or so. Thanks!

trevor-vincent:
Sorry for the delay on this, the official 2.7.0 release (https://github.com/taskflow/taskflow/archive/v2.7.0.tar.gz) compiles with the intel icpc compiler after a minor adjustment of the CMakeLists.txt as you suggested, however when compiling the master branch with icpc the build fails on the error:

```
[  3%] Building CXX object examples/CMakeFiles/condition.dir/condition.cpp.o
cd /scratch/p/pfeiffer/tvincent/taskflow/build/examples && /scinet/intel/2019u4/compilers_and_libraries_2019.4.243/linux/bin/intel64/icpc   -I/scratch/p/pfeiffer/tvincent/taskflow  -O3 -DNDEBUG   -std=gnu++17 -o CMakeFiles/condition.dir/condition.cpp.o -c /scratch/p/pfeiffer/tvincent/taskflow/examples/condition.cpp
In file included from /scratch/p/pfeiffer/tvincent/taskflow/taskflow/core/observer.hpp(3),
                 from /scratch/p/pfeiffer/tvincent/taskflow/taskflow/core/executor.hpp(3),
                 from /scratch/p/pfeiffer/tvincent/taskflow/taskflow/taskflow.hpp(3),
                 from /scratch/p/pfeiffer/tvincent/taskflow/examples/condition.cpp(20):
[...]

PR #251, opened 2020-12-01
Add support for the Intel C++ compiler

trevor-vincent:
This pull request is in response to issue #248. Versions of the intel C++ compiler icpc above v19.0.1 are known to support the C++17 core language features (see here https://en.cppreference.com/w/cpp/compiler_support). I have tested this pull request with versions v19.0.3.199 (2019-02-06)  and v19.1.3.304 (2020-09-25) of icpc.


tsung-wei-huang:
@trevor-vincent would you please fix the compile error :) There is a missing `;` after `_node->_handle.emplace<std::monostate>()`

trevor-vincent:
Thank you for fixing the ";" problem :)

tsung-wei-huang:
Thanks!!!

PR #257, opened 2020-12-12
Cancelling a taskflow

ojassm:
Sir,
Ojas Mithbavkar here. I had discussed with you a method involving a new function called _tear_cancelled_topology. You had suggested checking for the destruction of topology before completion of all references. It appears that such a case was happening in case of dynamic tasks. Hence I have devised a new method based on the join_counter.

In the normal run, the join_counter of the tpg keeps track of whether the node is last or not before tearing down the tpg. For cancelling, once the is_cancel flag is set to True, we prevent the execution of the function given to the node. 
```
if (!node->_topology->is_cancel){
    std::get<Node::StaticTask>(node->_handle).work();
  }
```

Next, we prevent any scheduling of its successors. 
```
if(--(node->_successors[i]->_join_counter) == 0) {
        if (!node->_topology->is_cancel){
          c.fetch_add(1);
          _schedule(node->_successors[i]);
        }
      }
```
If cancelled, the node is treated as a leaf node. The join_counter of the topology won't be incremented by the line `c.fetch_add(1);`  The counter would be decremented by `if (node->_topology->_join_counter.fetch_sub(1) == 1)` as always.  Thus, cancelling a tpg converts the existing nodes to leaf nodes, and the join_counter ensures that the tpg is torn down only once.
Also, to ensure that a false predicate does not reschedule a cancelled tpg, we add a check around that too.
```
if(! tpg->_pred() && !tpg->is_cancel  ) {
      assert(tpg->_join_counter == 0);
      tpg->_join_counter = tpg->_sources.size();
       _schedule(tpg->_sources); 
  }
```
The change in return type of exec.run from std::future to tf::Future remains.
Currently, this method passed all non-CUDA unit tests.

### DIFF
--- a/taskflow/core/executor.hpp
+++ b/taskflow/core/executor.hpp
@@ -338,9 +338,7 @@ auto Executor::async(F&& f, ArgsT&&... args) {
   using R = typename function_traits<F>::return_type;
 
   std::promise<R> p;
-  tf::Future<R> fu;
-  //auto fu = p.get_future();
-  fu.future_obj=p.get_future();
+  tf::Future<R> fu(p.get_future());
 
   auto node = node_pool.animate(
     std::in_place_type_t<Node::AsyncTask>{},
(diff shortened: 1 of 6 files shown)


tsung-wei-huang:
@ojassm thanks for the pull! I will review it and merge it then.

ojassm:
@tsung-wei-huang I have changed tf::Future to inherited class and removed the overloading methods.

ojassm:
@tsung-wei-huang One statement was missed in tf::Future due to which build had failed. I have resolved the error and the checks have passed.

PR #276, opened 2021-01-19
Expose Task hash value in TaskView class

Levi-Armstrong:


tsung-wei-huang:
@Levi-Armstrong thanks!

Issue #215, opened 2020-08-25
Windows Style Thread creation


mgood7123:
**Please describe your feature request.**

is it possible to implement windows style thread management into taskflow?

specifically the ability to create a thread in a suspended state, and thread_suspend/thread_resume

on UNIX this is achievable by using the clone(2) syscall in order to invoke pkill(..., SIG_STOP) without it also affecting the main program that manages the threads

for example

```C++
tf::Taskflow taskflow;

// Add tasks into the taskflow

// create A in a suspended state
tf::Task A = taskflow.emplace(
    [] () { std::cout << "Task A\n"; }
).suspend();
// can also suspend manually,
// suspending multiple times has no effect
// if task is already suspended
A.suspend();

// create B in a resumed state
tf::Task B = taskflow.emplace(
    [] () { std::cout << "Task B\n"; }
);

// Declare an executor
tf::Executor executor;

std::future<void> future = executor.run(taskflow);  // non-blocking return

// suspend task B if it has not finished executing yet
B.suspend();
// and resume task A
A.resume();

// do some stuff

// resume B
B.resume();

// wait for both A and B to end
future.wait();
```

tsung-wei-huang:
@mgood7123 this is definitely an interesting feature to do and I am very interested in getting this coroutine concept (or boost fiber) into Taskflow. It requires further discussion. For example, when you create a task, it doesn't really do anything but create a node that wraps the callable. I am not very clear what `A.suspend` means here? If you are interested in scheduling a call, please send us an email here: https://taskflow.github.io/

mgood7123:
> @mgood7123 this is definitely an interesting feature to do and I am very interested in getting this coroutine concept (or boost fiber) into Taskflow. It requires further discussion. For example, when you create a task, it doesn't really do anything but create a node that wraps the callable. I am not very clear what `A.suspend` means here? If you are interested in scheduling a call, please send us an email here: https://taskflow.github.io/

A.suspend() would simply suspend the task if it is running, or queue it for immediate suspension if it has not been executed or if it has finished via future.wait()

mgood7123:
or rather, it could set a flag telling the scheduler to not scedule the thread untill the flag is unset

tsung-wei-huang:
@mgood7123 , interesting - I believe this requires coroutine capability and another dedicated task type to handle (e.g., coroutine task).

mgood7123:
hmm ok

5fuheyan:
[...]

PR #279, opened 2021-01-21
can pause task impl

yangxingpping:
[https://github.com/taskflow/taskflow/issues/215#issue-685539450](url)

I am also want something like this. for example
A game instance[taskflow] compose of several components[task].
Task 1, create game instance; Task 2, deal players input; Task 3 clean up game instance(update player's level...)
And task 2 is driven by players' input, and May be looped for many time.


tsung-wei-huang:
@yangxingpping this seems interesting. Would you please try to help fix the compiler error of the CI so I can play around with it when it is done?

yangxingpping:
https://github.com/yangxingpping/taskflow/commit/c84324860bdc13c8150e57d347a43fb3b5f61299
Yes,  fixed.

tsung-wei-huang:
@yangxingpping can you explain the difference among pausable return, `true`, `false`, and `nullopt`?

yangxingpping:
true, resumeTask will call from paused task ( paused task will call again); false, resume Task will skip paused task; nullopt, task like staic static task

yangxingpping:
Does that can be achieved by other exists tasks?

tsung-wei-huang:
Not really - I was just a bit confused by the tri-state return :)

tsung-wei-huang:
@yangxingpping is it possible to do this at a taskflow scale? I am a bit skeptical about the practical use of pausing the graph from a task, especially when condition tasks come into play. However, being able to pause a submitted taskflow is more frequently used, based on my experience.

```cpp
tf::Taskflow taskflow;

// create some tasks

tf::Future<void> future = executor.run(taskflow);

// right, now we support cancel through the following:
// future.cancel();

// It is very useful if we can pause it as well
future.pause();
future.resume();
```

yangxingpping:
I want to have an method to control pause or continue taskflow accurately, so I think this should be done before call executor.run(taskflow).

junlinmessi:
What is the smallest granularity to pause? A Node?

tsung-wei-huang:
@yangxingpping thanks for your excellent effort. Your present implementation is cleaner than before (returning `std::optional<int>`). I will read the code as soon as possible and merge it to the branch.

tsung-wei-huang:
@yangxingpping Would you be able to provide some unittests for this new feature? I suggest creating a new file `unittests/pausable.cpp`. Also, `pausable` is probably better than `can pause`.

yangxingpping:
@tsung-wei-huang Yes, I will do in this week.

yangxingpping:
@junlinmessi smallest granularity is task.

PaiPratiksha:
does this PR handle "pause" Python Taskflow after a "failure" before "reverting"?

fran6co:
Why was this PR closed?

PR #298, opened 2021-03-08
fix template argument

cricketlong:
As we need to call `_device.read()`, the default template argument `Device` should be type of `std::istream`, I guess it was a copy-paste error.


tsung-wei-huang:
@cricketlong Yes, this is a copy-paste error. Thank you!

Issue #315, opened 2021-04-27
compile error with latest master + gcc 10


boxerab:
error: 

`/utility/serializer.hpp:469:18:   [ skipping 2 instantiation contexts, use -ftemplate-backtrace-limit=0 to disable ]`

in this code:

```
  if constexpr (std::is_arithmetic_v<typename U::value_type>) {
    _device.write(
      reinterpret_cast<const char*>(t.data()), 
      t.size() * sizeof(typename U::value_type)
    );
    sz += t.size() * sizeof(typename U::value_type);
  } else {
    for(auto&& item : t) {
     sz += _save(item);
    }
  }
```

this line: `    sz += _save(item);`

Full error log:

```
In file included from /home/aaron/src/grok/src/include/taskflow/core/graph.hpp:9,
                 from /home/aaron/src/grok/src/include/taskflow/core/task.hpp:3,
                 from /home/aaron/src/grok/src/include/taskflow/core/observer.hpp:3,
                 from /home/aaron/src/grok/src/include/taskflow/core/executor.hpp:3,
                 from /home/aaron/src/grok/src/include/taskflow/taskflow.hpp:3,
                 from /home/aaron/src/grok/src/lib/jp2/grk_includes.h:87,
                 from /home/aaron/src/grok/src/lib/jp2/t1/t1_part1/T1Part1.cpp:17:
/home/aaron/src/grok/src/include/taskflow/core/../utility/serializer.hpp: In instantiation of ‘SizeType tf::Serializer<Device, SizeType>::_save(T&&) [with T = const std::__cxx11::basic_string<char>&; std::enable_if_t<is_std_basic_string_v<typename std::decay<_Up>::type>, void>* <anonymous> = 0; Device = std::basic_ofstream<char>; SizeType = long int]’:
/home/aaron/src/grok/src/include/taskflow/core/../utility/serializer.hpp:422:16:   required from ‘SizeType tf::Serializer<Device, SizeType>::operator()(T&& ...) [with T = {const std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&, const tf::TaskType&, const std::chrono::time_point<std::chrono::_V2::steady_clock, std::chrono::duration<long int, std::ratio<1, 1000000000> > >&, const std::chrono::time_point<std::chrono::_V2::steady_clock, std::chrono::duration<long int, std::ratio<1, 1000000000> > >&}; Device = std::basic_ofstream<char>; SizeType = long int]’
[...]

PR #316, opened 2021-04-27
warnings: add a few static_casts to fix warnings

boxerab:
Hi @tsung-wei-huang ,
These warning fixes also fix the error I was seeing in #315,
Cheers,
Aaron


boxerab:
Simpler for me to just wrap the taskflow header in GCC pragmas to disable sign conversion warnings. But, I still think these warnings should be carefully checked to make sure they are not hiding real bugs.

Issue #303, opened 2021-03-20
Better / more control over threads


jonesmz:
Related to :
https://github.com/taskflow/taskflow/issues/191
https://github.com/taskflow/taskflow/issues/164
https://github.com/taskflow/taskflow/issues/190
https://github.com/taskflow/taskflow/issues/204

I want better control over the treads used to execute task flow tasks.

In my situation. I want to be able to call `tf::WorkerThread::run(executor)` in any arbitrary thread (but not inside the callstack of a previous call to `tf::WorkerThread::run()`!) To make the current thread execute jobs. The call to `run()` would only return when either the executor is shutting down, or `tf::WorkerThread::cancel()` is called with the current threads worker id.

This solves the following situations:

- Resources that can only be accessed / managed from a thread that is not owned by tf::Executor, such as opengl resources or other 3rd party frameworks.
- Micromanaging properties of the threads for an executor, such as assigning OS scheduler priorities. Numa nodes, security considerations such as setting system call restrictions on the worker threads, and whatever else someone might want to do.
- Multitasking systems where the available resources for a particular task flow need to change in response to external constraints, such as a new job being submitted that has a higher priority. The job manager could remove one or more threads from the current task flow and attach them to the new job's task flow instead. And migrate them back when the new job is finished.
- The "single threaded" mode of execution, where a task flow is created, and then used in the same thread it was created in. Previously this was an executor with 0 threads.
- Makes it easier to avoid over-threading an application. While its true that using a few more threads than the CPU supports can provide slightly higher density than if you had only threads equal to numcpu. Due to some threads occasionally blocking on system calls and such. The current model of `tf::Executor`  makes it far too easy to overcommit by creating multiple different executors and using them concurrently with each other. Or similarly bad: Creating new `tf::Executor`s as new task flows are constructed, executing the taskflow and then shutting down the executor. This makes the end user pay the cost of spawning new threads every time.

Looking at the code for Executor. I don't see this as being a very large change internally. And the current model could continue being supported as well, with the same API. On top of a base class that provided "bring your own threads" support.

The way threads are managed now is to simply spawn new `std::thread`s with the construction of the `tf::Executor`. And have them call the `_exploit_task` function. While I do see occasional implicit reliance on the N variable to indicate the number of worker threads that exist, that's simple enough to replace with a call to `workers.size()`

[...]

PR #323, opened 2021-05-26
Thread-agnostic scheduling

z-adams:
This is a sketch of a solution to #303. I noticed that you guys are actively working on the same code I've modified here (for instance, dynamically calculating `max_steals` in `Executor::_explore_task`) and I've torn a lot of stuff up here so I don't actually expect this code to be merged, but hopefully it will provide some inspiration and generate some discussion.

Unfortunately the diff is pretty unusable because of the way I've shuffled the files around but I'll summarize the main changes:

* `class Executor` was turned into a thin wrapper for a base class called `class TaskScheduler` which stores the previous contents of `class Executor`. `Executor`'s job is now to simply store a `std::vector<std::thread>`, so that `TaskScheduler` no longer has to be explicitly tied to its executing threads.
* `TaskScheduler` no longer has any explicit dependence on `N`. Instead, threads are registered as workers with `TaskScheduler::register_worker(std::thread&)`, which configures the thread and adds a new worker to the vector of workers. 
* When all the threads are registered, `TaskScheduler::_configure()` is called to set up the things that were previously done in the constructor based on the number of workers `N`. I've used placement `new` to re-initialize the `Notifier` which is pretty ugly, someone who understands the codebase better may have a better idea.
* I also added a `TaskScheduler::_shutdown()` function, so that task execution can be halted before the wrapper class `join()`s its member threads.

The interface I've created for `TaskScheduler` doesn't have to actually be this way, it's just what I came up with on my way to turning the execution code into a usable base class.

This doesn't solve some of the other common requests related to #303 such as allowing the main application thread to be a worker, or requiring certain tasks to execute on particular threads, or adding or removing threads during execution, but it's an important step in decoupling the details of the worker from the executor, and does solve the issues with over-threading.


taskflow/core/task_scheduler.hpp
```diff
@@ -0,0 +1,1270 @@
...
+    /**
+    @brief configures the scheduler from the current workers
+    */
+    void _configure();
+
+    /**
+    @brief shuts down the scheduler; must be called for destruction
+    */
+    void _shutdown();
+
```
z-adams, line 196: Interface to allow management by wrapper

taskflow/core/task_scheduler.hpp
```diff
@@ -0,0 +1,1270 @@
...
+{
+  size_t N = _workers.size();
+      
+  if(N == 0) {
+    TF_THROW("no cpu workers to execute taskflows");
+  }
+
+  // Re-construct Notifier with placement new (it has no copy constructor)
+  _notifier.~Notifier();
+  new (&_notifier) Notifier(N);
```
z-adams, line 272: This is ugly but Notifier isn't easy to dynamically update, or even reinitialize since its copy constructor is implicitly deleted due to its `std::atomic<>` member field

tsung-wei-huang:
@z-adams thank you for the pull request. I will take a look at it and see how it works.

PR #324, opened 2021-05-28
Update notifier.hpp

MambaWong:


tsung-wei-huang:
What is this change?

MambaWong:
I think the _increase_ is easier to understand using _shift._
https://github.com/taskflow/taskflow/blob/9415c3b68fd050889f57412394933e94b74755e0/taskflow/core/notifier.hpp#L235

Issue #328, opened 2021-06-23
How to create a dynamic subflow from within an  for_each_index operation ie: nested for_each_index.


bradphelan:
The example given in the docs is

```
taskflow.for_each_index(0, 100,  2, [](int i) { });  // 50 loops with a + step
```

which represents

```
for(size_t i = 0; i<100;i+=1){}
```

but how to do a nested loop given that the callback for ``for_each_index`` does not pass a subflow object in?

```
for(size_t i = 0; i<5;i+=1){
    for(size_t j = 0; i < 100; j+=1){
        someTask(i,j);
    }
}
```

bradphelan:
I think I've figured out what the detail is in the implementation. for_each_index uses async_silent for each subtask and this is always a leaf node in the task system. You can't spawn subflows from them I think? I guess I would be looking for something like

```
taskflow.for_each_index_sf(0, 100,  2, [](int i, tf::Subflow & sf) { 
    sf.for_each_index(0, 100,  2, [i](int j) { 
       std::cerr << i << j << std::endl;
    });
});  
```

Is this reasonable or am I way off track?

bradphelan:
Ok I got this to work

```
TEST(TASK_FLOW,nested_for) {
	auto N = 10;

	if (N < 0)
	{
		throw std::runtime_error("N must be non-negative");
	}

	int res;  // result

	tf::Executor executor;
	tf::Taskflow taskflow("nestedfor");

	taskflow.for_each_index_sf(0,5,1,[](int i, tf::Subflow&sfi){
		sfi.for_each_index(25,30,1,[i](int j){
			std::cerr << i << " "  << j << std::endl;
		});
	});

	executor.run(taskflow).wait();


}
```

using an implementation of

```
template <typename B, typename E, typename S, typename C>
Task FlowBuilder::for_each_index_sf(B&& beg, E&& end, S&& inc, C&& c){
  
  using I = stateful_index_t<B, E, S>;
  using namespace std::string_literals;

  Task task = emplace(
  [b=std::forward<B>(beg), 
   e=std::forward<E>(end), 
   a=std::forward<S>(inc), 
   c=std::forward<C>(c)] (Subflow& sf) mutable {
    
    // fetch the iterator values
    I beg = b;
    I end = e;
    I inc = a;

    if(is_range_invalid(beg, end, inc)) {
      TF_THROW("invalid range [", beg, ", ", end, ") with step size ", inc);
    }
    
    size_t chunk_size = 1;
[...]

PR #329, opened 2021-06-24
Added for_each_index_nested ( do not merge yet )

bradphelan:
Pull request for #328 

I've added in for_each_index_nested however I cannot get the the test case to work. Have a look at the test case in

unittests/algorithm.cpp

if I use just for_each_index then the test case passes. If I use the implementation of for_each_index_nested as you reviewed yesterday then there is an error. Would you please have a look at and see what the problem may be.

### DIFF
--- a/examples/parallel_for.cpp
+++ b/examples/parallel_for.cpp
@@ -34,6 +34,35 @@ void for_each_index(int N) {
   executor.run(taskflow).get();
 }
 
+
+void for_each_index_nested() {
+	auto N = 10;
+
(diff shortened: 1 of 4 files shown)


XGZepto:
Hi Brad,
came across this PR since it would be super helpful for the project I'm working out. I'll take a look at your code but do let us know if there's any update from your side since the PR was opened almost 3 years ago. Cheers.

bradphelan:
No we never updated this

Issue #337, opened 2021-08-31
Crash using FlowBuilder::sort due to corruption of 'cmp' stored on stack


fstrugar:
Hi, 

I stumbled upon what looks like a bug using FlowBuilder::sort - got a crash that looked like it was caused by memory corruption of my compare lambda capture members. Not observable when custom compare lambda has no captures (as is the case in examples) and not observable with smaller number of elements to sort (i.e. below 10,000).

I don't have an easy repro (I can build it) but I think I tracked it down to this line:
https://github.com/taskflow/taskflow/blob/master/taskflow/core/algorithm/sort.hpp#L346

Capturing 'comp' as a reference will capture it from the current stack but since it's an async call, it can get executed after the 'parent' function ends, thus getting a corrupt stack.

So just changing to capture it as a value fixes my crash. I can build a repro if needed but I think capture by reference should be incorrect anyway?

fstrugar:
Added pull request with a fix: https://github.com/taskflow/taskflow/pull/338

PR #338, opened 2021-08-31
Fix for FlowBuilder::sort due to by ref lambda capture of comparator function (issue #337)

fstrugar:
parallel_pdqsort's recursive async lambda call to parallel_pdqsort was capturing a reference to Compare on caller function stack which isn't guaranteed to remain valid due to nature of Subflow::silent_async call; made it capture by value instead

https://github.com/taskflow/taskflow/issues/337


tsung-wei-huang:
@fstrugar thank you for the catch!

PR #340, opened 2021-09-02
Enable users to associate arbitrary data with a Task

### LINKED ISSUE #335: Attaching user data to tasks
Is it possible or considered for future development to attach user data to a task? My primary interest is a seemingly simple case when a taskflow consists of tasks each with consistent set of boolean flags attached to it. They are used by conditional tasks to skip over some calculations if certain flags are set in dependencies.

It seems that `Node::name` is similar kind of data. Maybe it could be generalized for such use cases.

I can also help implementing this.

Endilll:
Closes #335.

### DIFF
--- a/taskflow/core/graph.hpp
+++ b/taskflow/core/graph.hpp
@@ -207,10 +207,12 @@ class Node {
     size_t num_weak_dependents() const;
 
     const std::string& name() const;
+    void* data() const;
 
   private:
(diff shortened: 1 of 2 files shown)


tsung-wei-huang:
@Endilll thank you! I will merge it later today after I fixed the tsan issue on the main project.

PR #342, opened 2021-09-06
Add erase to taskflow and graph

Endilll:
I decided to add `Graph::erase` with `node_pool.recycle()` invocation inside to mirror `Graph::emplace` which invokes `node_pool.animate()`.


tsung-wei-huang:
@Endilll thank you for the contribution. Yes, this is the correct implementation.

PR #350, opened 2021-09-28
Inrease macOS compatibility

burgholzer:
This PR aims to increase the compatibility of the taskflow library with older versions of macOS.
Trying to build Python wheels for our project that incorporates taskflow showed that targeting macOS yields several errors stating that std::variant's `std::get` is not available.
For an example of this see https://github.com/iic-jku/ddsim/runs/3732494041?check_suite_focus=true#step:4:280.

The problem can be resolved by replacing `std::get` calls with the corresponding `std::get_if` equivalent (see also https://stackoverflow.com/a/53887048).

### DIFF
--- a/taskflow/core/executor.hpp
+++ b/taskflow/core/executor.hpp
@@ -918,7 +918,7 @@ inline void Executor::_tear_down_invoke(Node* node, bool cancel) {
     // async task needs to carry out the promise
     case Node::ASYNC:
       if(cancel) {
-        std::get<Node::Async>(node->_handle).work(true);
+          std::get_if<Node::Async>(&(node->_handle))->work(true);
       }
       _tear_down_async(node);
     break;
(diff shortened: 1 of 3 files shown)


tsung-wei-huang:
@burgholzer Thanks for the pull - this is MAC OS's issue which I am aware of. I am not a big fan of changing this to `get_if` because the extra `if` block does need an additional comparison. However, I will see what I can do with your pull request.

BTW, are you using taskflow in your quantum simulator project? I know Robert in JKU.  If yes, I am interested in learning more about your computational patterns and maybe we can collaborate on something. You may shoot me an email: tsung-wei.huang@utah.edu

burgholzer:
Thanks for the quick reply! As far as I understand it, both, `std::get` and `std::get_if` have to perform an additional check. While the former has to check whether the type of the object matches the template parameter and throwing an exception if not, the latter compares both types and returns a `nullptr` whenever a mismatch occurs. So it just seems to be a trade off between a construct that possibly throws an exception and one that does not.
In addition some compilers may even be able to optimize that check away due to the `index` check beforehand. Although this would need some testing.

Regarding your question: Yes, we are trying to use taskflow in our decision diagram-based quantum simulator to allow for parallelization and utilizing more degrees of freedom. I'll send you an email with further details 👍🏼

tsung-wei-huang:
@burgholzer thank you. Yes, it's correct. In this case, `std::get_if` is actually better due to `noexcept`. I have merged your pull.

I am very interested in learning more about your simulator and maybe we can come up with something using GPU to speed up the simulation.

burgholzer:
Many thanks for resolving this so quickly. I'll reach out to you soon!

PR #362, opened 2021-12-07
Test non-random accessible iterators

antonysigma:
In addition to testing tf::transform with `std::vector<T>` iterators,
we also test containers without fast random access iterators, e.g.
`std::list<T>`.

### DIFF
--- a/unittests/algorithm.cpp
+++ b/unittests/algorithm.cpp
@@ -1098,29 +1098,30 @@ TEST_CASE("ps.object.4.100000") {
 // parallel transform
 // ----------------------------------------------------------------------------
 
+template<class T>
 void parallel_transform(size_t W) {
-  
+
   std::srand(static_cast<unsigned int>(time(NULL)));
-  
+
   tf::Taskflow taskflow;
   tf::Executor executor(W);
 
   for(size_t N=0; N<1000; N++) {
 
-    std::vector<int>::iterator src_beg;
-    std::vector<int>::iterator src_end;
+    typename T::const_iterator src_beg;
+    typename T::const_iterator src_end;
     std::list<std::string>::iterator tgt_beg;
 
-    std::vector<int> src;
+    T src;
     std::list<std::string> tgt;
-    
+
     taskflow.clear();
 
     auto from = taskflow.emplace([&](){
       src.resize(N);
       for(auto& d : src) {
         d = ::rand() % 10;
-        tgt.push_back("hi");
+        tgt.emplace_back("hi");
       }
       src_beg = src.begin();
       src_end = src.end();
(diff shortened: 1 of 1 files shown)


tsung-wei-huang:
Feel free to add subflow test for parallel-transform and turn this pull to a ready status for merge.

unittests/algorithm.cpp
```diff
@@ -1213,17 +1219,117 @@ void parallel_transform2(size_t W) {
...
+      const auto [src_beg, src_end] = src.equal_range(0);
+      const size_t n_matching = std::distance(src_beg, src_end);
+      tgt.resize(n_matching);
+
+      subflow.transform(std::ref(src_beg), std::ref(src_end), std::ref(tgt.begin()),
+        [&] (const auto& x) -> string {
+          return myFunction(x.second);
+      });
+    });
+
```
antonysigma, line 1306: Hi @tsung-wei-huang , I turned this PR from "draft" to "ready" as you instructed.

I am also half way through hammering out the test case for `Subflow::transform`. Currently the compiler complains about this error. I will debug my test code later this evening.

```c++
FAILED: unittests/CMakeFiles/algorithm.dir/algorithm.cpp.o 
/usr/bin/c++  -I/home/antony/Documents/taskflow -I/home/antony/Documents/taskflow/3rd-party/doctest -O3 -DNDEBUG -Wall -Wextra -Wfatal-errors -std=gnu++1z -MD -MT unittests/CMakeFiles/algorithm.dir/algorithm.cpp.o -MF unittests/CMakeFiles/algorithm.dir/algorithm.cpp.o.d -o unittests/CMakeFiles/algorithm.dir/algorithm.cpp.o -c /home/antony/Documents/taskflow/unittests/algorithm.cpp
/home/antony/Documents/taskflow/unittests/algorithm.cpp: In lambda function:
/home/antony/Documents/taskflow/unittests/algorithm.cpp:1301:87: error: use of deleted function ‘void std::ref(const _Tp&&) [with _Tp = __gnu_cxx::__normal_iterator<std::__cxx11::basic_string<char>*, std::vector<std::__cxx11::basic_string<char> > >]’
       subflow.transform(std::ref(src_beg), std::ref(src_end), std::ref(tgt.begin()),
                                                                                   ^
```

tsung-wei-huang: Thanks. I don't think you can `std::ref` a r-value (`tgt.begin()`), which may be the reason of compiler error.

antonysigma: @tsung-wei-huang I finished debugging the test code. Could you please review the code again? 

`cmake -GNinja . build/ && ninja -C build/ test` returns 100% test pass.

tsung-wei-huang:
Thanks for contributing to the feature and its test!

tsung-wei-huang:
I added `subflow.join` as you will need those iterators valid during the context.

antonysigma:
> I added `subflow.join` as you will need those iterators valid during the context.

Sure. Thanks for the tip.

PR #374, opened 2022-01-08
Update cuda_meta.hpp

hjxy2012:
removed std::binary_function.


tsung-wei-huang:
I don't think it is necessary to inherit `T` (`public T`)?

hjxy2012:
> 

Thank you for your reply. I just removed public inheritance of T.

PR #383, opened 2022-02-28
examples: add benchmark comparing taskflow to good old thread pool

boxerab:
I thought it would be useful to compare taskflow performance to a simple thread pool

Also, I notice that the performance of taskflow drops relative to the thread pool 
as the number of iterations increases. Any comments on why this is the case ?

Thanks!


tsung-wei-huang:
@boxerab thanks for this pull request! would you mind migrating it to the `benchmarks/` folder instead of the `examples/`?

boxerab:
@tsung-wei-huang done. Just want to say thank for this fantastic library.

tsung-wei-huang:
This may not be important to this pull (and I will change it later), yet you can reuse the same `tf::Taskflow` object in the inner loop:

```
tf::Taskflow taskflow;
for(int i=0; i<N; i++) {
  taskflow.clear();
  // ... add tasks
}
```

A `tf::Taskflow` object can be reused which allows more efficient object pooling under the hood.

boxerab:
> This may not be important to this pull (and I will change it later), yet you can reuse the same `tf::Taskflow` object in the inner loop:
> 
> ```
> tf::Taskflow taskflow;
> for(int i=0; i<N; i++) {
>   taskflow.clear();
>   // ... add tasks
>   executor.run(taskflow).wait();
> }
> ```
> 
> A `tf::Taskflow` object can be reused which allows more efficient object pooling under the hood.

Thanks, I wasn't calling clear in the original implementation. I have to admit I was a little concerned when taskflow was 10x slower  :)  All good now.

PR #395, opened 2022-04-16
disable installation for subprojects

maxbachmann:
This disables the installation of Taskflow when it is used via `add_subdirectory`. This is e.g. the case when using it with `FetchContent_MakeAvailable`.


maxbachmann:
Probably it makes sense to disable e.g. the Tests when using Taskflow as submodule as well.

PR #410, opened 2022-06-17
Add ability to set max steals and max yields

mpowelson:
When profiling our application that uses Taskflow, we noticed that `std::this_thread::yield();` was causing a lot of cpu spin time. After discussing with @tsung-wei-huang, it sounds like the main problem was that our system uses multiple executors in order to keep resources independent for different types of tasks. He suggested that we try to reduce the number of executors and also that we adjust the max steals and max yields. After testing reducing the max yields and max steals (first 2 commits), this didn't seem to have much of an impact on spin time. As a result, I added an optional sleep after max yields is reached (3rd commit). This does reduce the spin time.

This PR adds the ability to set the max steals, max yields, and the sleep duration after those are reached. The defaults should work the same as before if you were using all default arguments for the executor. If you were passing in the number of threads (N), simply pass in ((N + 1) << 1) for max steals to maintain the same behavior.

### DIFF
--- a/taskflow/core/executor.hpp
+++ b/taskflow/core/executor.hpp
@@ -65,12 +65,14 @@ class Executor {
     hardware concurrency returned by std::thread::hardware_concurrency.
     @param max_steals Sets _MAX_STEALS. Defaults to ((std::thread::hardware_concurrency() + 1) << 1)
     @param max_yields Sets _MAX_YIELDS. Defaults to 100
+    @param max_yields_sleep_time Sets _MAX_YIELDS_SLEEP_TIME. Defaults to 0us
     */
     explicit Executor(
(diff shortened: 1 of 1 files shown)


tsung-wei-huang:
Hi @mpowelson ,

Sorry for the late response. It has taken me a while to think of a better solution for avoiding this issue. There is recently a pull request #440 about solving the problem. I have designed a new invariant based on the discussion with @p12tic in that pull request. It is placed in the main branch now, and hopefully will be merged to the next release. Please test it again on your workload and see if the problem still exists.

tsung-wei-huang:
This is perfect! Thank you very much. @p12tic might also be interested in having a look at this result.

PR #428, opened 2022-08-18
fix compile command

guodongxiaren:
because there is`#include <taskflow/taskflow.hpp>`  in your simple.cpp, g++ doesn't need to specify another taskflow directory as -I parameter.
for example, when I run:
```bash
g++ -std=c++17 simple.cpp -I taskflow/taskflow -O2 -pthread -o simple
```
I receive this error message:
```
simple.cpp:1:10: fatal error: taskflow/taskflow.hpp: No such file or directory
    1 | #include <taskflow/taskflow.hpp>  // Taskflow is header-only
      |
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ tell the compiler to include the [headers](./taskflow/).
 
 ```bash
 ~$ git clone https://github.com/taskflow/taskflow.git  # clone it only once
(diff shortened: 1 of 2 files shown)


tsung-wei-huang:
Thanks!

PR #434, opened 2022-09-13
Resolved MSVC warning C4456

matt77hias:
Resolved MSVC warning C4456: declaration of 'lock' hides previous local declaration


tsung-wei-huang:
Thank you for the pull! This has been fixed.

PR #442, opened 2022-11-03
Improve inlining of critical functions of TaskQueue

p12tic:
`TaskQueue::push()` and `TaskQueue::pop()` are functions that are in critical hot code path and used only from few places. It makes sense to look into improving inlining decisions for these functions.

As compiled on x86_64 by gcc 10.2, `TaskQueue::push()` contains a number of function prologue and epilogue instructions. These are mostly needed for the case when the queue overflows and needs resizing, which happens extremely infrequently. Inlining the frequent code path and forbidding inlining of the infrequent code path should improve performance slightly.

`TaskQueue::pop()` on the other hand does not have much in function prologue and epilogue as it fits into caller-saved registers. However, the function is called from only 2 places, so inlining it too does not increase binary size too much yet should allow better register usage in the caller.

Preliminary testing shows that these optimizations lead to up to 0.5% faster benchmarks (depends on benchmark). I will provide full result spreadsheet when I run the benchmarks enough times.

### DIFF
--- a/taskflow/core/tsq.hpp
+++ b/taskflow/core/tsq.hpp
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "../utility/macros.hpp"
 #include "../utility/traits.hpp"
 
 /**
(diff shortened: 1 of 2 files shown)


tsung-wei-huang:
@p12tic thank you for the pull request! Yes, pop and push are on the hot path of work-stealing.

p12tic:
@tsung-wei-huang I've setup some counters and the frequency of `_notifier.notify()` calls from `_schedule()` are minuscule as implemented in #440. It's on the order of less than ten per second.

PR #444, opened 2022-11-08
Implement fallback uncaught exception handler

p12tic:
Currently Taskflow is not exception safe as exceptions thrown from user tasks will break invariants required by the library and most likely result in deadlocks. This situation can be improved by providing a fallback exception handler which is called for any exceptions originating from user tasks. This way exceptions are processed in a well-defined way.

The API explicitly disclaims guarantee to provide the same exact way to handle exceptions in future versions of the library, so this change does not reduce freedom to improve exception handling in the future.

This PR introduces two simple tests with tasks that throw exceptions. Both tests result in a deadlock with code before this PR.

### DIFF
--- a/taskflow/core/executor.hpp
+++ b/taskflow/core/executor.hpp
@@ -702,6 +702,26 @@ class Executor {
     */
     size_t max_steals() const noexcept;
 
+    /**
+    @brief sets handler for exceptions that are propagated from
+    user-provided task callables and for which the library does not
+    specify any other way of handling them.
+
+    There is no synchronization when invoking the handler.
+    If exception propagates out of handler it is ignored.
+
+    This exception handler is considered a fallback to handle exceptions.
+    There is no guarantee that exceptions originating from a particular
+    task will be delivered to the handler passed to this function in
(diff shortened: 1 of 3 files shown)


p12tic:
@tsung-wei-huang Just a friendly ping.

tsung-wei-huang:
@p12tic Thanks for the pull request - and sorry for the delayed response. I was on Thanksgiving break :) Here are some points I think:

The default uncaught exception handler should be more informative. For example, instead of making the callback `std::function<void(std::exception_ptr)>` we may consider an extensible interface `std::function<void(tf::Exception&)>` where `tf::Exception` can include more information (e.g., which worker, which task, which taskflow, etc.) for users to probe.

```
struct ExceptionHandle {
  std::exception_ptr exception_ptr;
  Worker& worker;
  Task task;
};
```

What do you think?

bradphelan:
Has this progressed?

tsung-wei-huang:
close for now as the current branch incorporated support for exception :) Will be part of the next release 3.7.0

PR #448, opened 2022-11-12
Update README.md

mconcas:
Hi, I'm new to taskflow, while I was running the basic tutorial following the readme, I saw this small glitch.
In case, the proposed command worked for me.

Keep up with the nice work, cheers.


tsung-wei-huang:
Thanks for the pull! Yes, it was a glitch I didn't notice :)

PR #452, opened 2023-01-02
Define num_blocks

bkmgit:
Prevent compilation error when building wavefront example.


tsung-wei-huang:
Thank you for the fix!

PR #454, opened 2023-01-11
Update cmake minimum version to 3.18

BeingGod:
Current cmake minimum version is set to cmake 3.12 in CMakeLists.txt. This version allows to define `CMAKE_CUDA_STANDARD` up to 14. Support for `CMAKE_CUDA_STANDARD` 17 and 20 have only been introduced recently in cmake 3.18.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,7 +10,7 @@
 
 
 # CMake version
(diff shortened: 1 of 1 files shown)


tsung-wei-huang:
Thank you for the fix!

PR #459, opened 2023-02-24
Renamed MAX_PRIORITY to TF_MAX_PRIORITY

Notallthatevil:
winspool.h has a define MAX_PRIORITY. If that header is included somewhere along the lines along with taskflow.hpp then build errors occur. Simply prepending TF_ to the MAX_PRIORITY template parameter resolves this collision.


tsung-wei-huang:
Thanks!!!

PR #473, opened 2023-04-01
Adding STL algorithms inclusive_scan and exclusive_scan

reachomk:
Hi!

As I mentioned I would in my GSoC proposal, I've implemented parallelized versions of std::inclusive_scan and std::exclusive_scan. I've organized these such that you can simply include <taskflow/algorithms/numeric.hpp>, similar to how one would just include <numeric> for these algorithms. 

Let me know what you think!


tsung-wei-huang:
Thanks for your pull request. I don't think the simple `for_each_index` can work for the scan - it needs a cumulative scan to find the summation. I have implemented a correct version which you can take a look at [here](https://github.com/taskflow/taskflow/blob/master/taskflow/algorithm/scan.hpp). Of course, the performance may be further fine-tuned.

reachomk:
Hi Dr Huang, 

Thanks so much for putting in the time to implement these. I just glanced at the code and had 2 questions.

1. Could you please clarify what you mean by cumulative scan, and how that differs from my implementation?

2. For inclusive_scan, you create a task using the emplace function, and within that, you use a for loop to iterate through the number of workers. Why do we use a for loop? To the best of my knowledge, for loops don't run in parallel, so wouldn't we use for_each_index for this?

PR #486, opened 2023-06-04
Update compatibility guarantees

maxbachmann:
There are breaking changes in many minor Version upgrades, so the compatibility is really only between minor versions.

E.g. Version 3.6 introduced a couple oft breaking changes leading to build errors downstream https://github.com/maxbachmann/RapidFuzz/issues/329


tsung-wei-huang:
Thanks for the fix!

PR #496, opened 2023-08-22
Add a function to cancel the precedence relationship between two tasks.

QiuYilin:
I am writing a data flow gui. The bottom layer of each node corresponds to a task in taskflow. I want to map the connection and disconnection operations of the nodes on the gui to tasks in taskflow. Because taskflow already has the precede and succeed interfaces corresponding to the connection operation, I add a function called `void connectDecre(Task dep, Task cuc) ` which is implemented to cancel the connection. 
I don't know if this function is necessary at the level of the entire project. If it doesn't meet the design intent and this pull request is unreasonable, I may need to separate my own data flow structure of the gui from the taskflow structure. The data flow structure of gui  can be dynamically operated, and then it is parsed into taskflow when it is finally executed. Or I just use inheritance method or additional code to meet my own needs.

### DIFF
--- a/taskflow/core/taskflow.hpp
+++ b/taskflow/core/taskflow.hpp
@@ -255,6 +255,13 @@ class Taskflow : public FlowBuilder {
     */
     Graph& graph();
 
+    /**
+    @brief 
+
+    Cancel the precedence relationship between two tasks.
+    */
+    void connectDecre(Task dep, Task cuc);
+
   private:
 
     mutable std::mutex _mutex;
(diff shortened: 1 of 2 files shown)


tsung-wei-huang:
what about renaming it to "erase_dependency"?

tsung-wei-huang:
@QiuYilin Thank you for the contribution. I have added a new method `Taskflow::remove_dependency` available in the current main branch.

QiuYilin:
> @QiuYilin Thank you for the contribution. I have added a new method `Taskflow::remove_dependency` available in the current main branch.

Thank you. Taskflow::remove_dependency is a better name.

PR #505, opened 2023-09-12
Add task_wrapper for for_each, for_each_index, transform and find_if …

### LINKED ISSUE #504: Per chunk RAII support
**Issue Summary:**

We are encountering a problem related to managing floating point exceptions (FPU) on the CPU. Currently, we are using a workaround involving a lambda function to maintain FPU settings across subtasks, but this approach is both costly and inefficient. Instead, we need to implement a default chunk wrapper that applies FPU settings at the start of a task and restores the previous FPU state at the task's end. This change will optimize performance and prevent corruption of tasks that yield to the executor.

**Problem Description:**

We need to manage the FPU settings for CPU operations in our code. The challenge is ensuring that all subtasks created after setting the FPU context inherit the parent task's FPU settings. Currently, we achieve this by encapsulating the loop body within a lambda function. This lambda function captures the FPU state before calling `executor.corun` and then applies this FPU state on every iteration of the loop. However, this approach has proven to be inefficient and resource-intensive.

The inefficiency arises from the overhead associated with making these FPU state calls on every iteration of the loop. Moreover, this approach is unnecessary as FPU settings only need to be applied at the beginning of each chunk or task. At the end of each task, we must also restore the previous FPU state to avoid corrupting any tasks that have yielded to the executor at this point.

**Proposed Solution:**

To address this problem efficiently, we propose implementing a default chunk wrapper for all algorithms. This wrapper would take care of applying the necessary FPU settings at the beginning of each chunk or task, and it would ensure the restoration of the previous FPU state at the task's end. By doing so, we can eliminate the overhead and inefficiency associated with applying FPU settings on every loop iteration, resulting in improved performance and avoiding task corruption issues.

**Desired Outcome:**

Implement a default chunk wrapper that handles FPU settings at the start and end of tasks, optimizing performance and preventing task corruption issues. This change will simplify code and improve overall efficiency in managing floating point exceptions on the CPU.

**Psuedo Code**

```
auto defaultChunkWrapper = [](auto && chunkRunner){
    std::forward<decltype(chunkRunner)>(chunkRunner());
};
```

however in our case we want something like

```
FPUSaver saver;
auto fpuChunkWrapper = [&saver](auto && chunkRunner){
    // Appy stored FPU settings to the current tasks and
    // restore the FPU setting as the end of the task
    FPUApplier applier(saver);
    std::forward<decltype(chunkRunner)>(chunkRunner());
};

Taskflow taskflow;
taskflow.for_each_index(0,10000,1,[](double i){  
     // trigger FP exception
     i / 0.0;
}, tf::DynamicPartioner(), chunkRunner);

executor.corun(taskflow);
```

The global https://taskflow.github.io/taskflow/classtf_1_1ObserverInterface.html interface doesn't really solve the problem in an elegant way.

bradphelan:
fixes #504 

This can be used for multiple purposes. For handling exceptions we could do

```cpp
tf::Executor executor(tc);
std::exception_ptr x;
auto task_wrapper = [&](auto &&task)
{
  try {
  task();
  }catch(...){
    x = std::current_exception();
  }
};
std::atomic<int> count = 0;
tf::Taskflow taskflow;
taskflow.for_each_index(
    0, 100, 1, [&](int i)
    { count++; },
    tf::StaticPartitioner(1), task_wrapper);
if(x)
   std::rethrow_exception(exceptionPtr);
```

Our use case is to apply captured floating point settings from parent tasks to sub tasks 

```
class FPUSaver;
class FPUApplier;
```

```cpp
tf::Executor executor(tc);
std::exception_ptr x;
/// Capture the floating point unit configuration
FPUSaver saver;
auto task_wrapper = [&](auto &&task)
{
  try {
    /// Apply the floating point unit configuration ( and undo on destruction
    FPUApplier applier(saver);
    task();
  }catch(...){
    x = std::current_exception();
  }
};
std::atomic<int> count = 0;
tf::Taskflow taskflow;
taskflow.for_each_index(
    0, 100, 1, [&](int i)
    { count++; },
    tf::StaticPartitioner(1), task_wrapper);
if(x)
   std::rethrow_exception(exceptionPtr);

```

### DIFF
--- a/taskflow/algorithm/for_each.hpp
+++ b/taskflow/algorithm/for_each.hpp
@@ -1,18 +1,20 @@
 #pragma once
 
 #include "launch.hpp"
+#include "../core/task_wrapper.hpp"
 
 namespace tf {
 
 // Function: make_for_each_task
-template <typename B, typename E, typename C, typename P = GuidedPartitioner>
-TF_FORCE_INLINE auto make_for_each_task(B b, E e, C c, P&& part = P()) {
-  
+template <typename B, typename E, typename C, typename P = GuidedPartitioner, typename TW = TaskWrapperIdent>
+TF_FORCE_INLINE auto make_for_each_task(B b, E e, C c, P &&part = P(), TW &&task_wrapper = TW())
+{
+
   using B_t = std::decay_t<unwrap_ref_decay_t<B>>;
   using E_t = std::decay_t<unwrap_ref_decay_t<E>>;
   using namespace std::string_literals;
 
-  return [b, e, c, part=std::forward<P>(part)] (Runtime& rt) mutable {
+  return [b, e, c, task_wrapper, part=std::forward<P>(part)] (Runtime& rt) mutable {
 
     // fetch the stateful values
     B_t beg = b;
@@ -23,7 +25,9 @@ TF_FORCE_INLINE auto make_for_each_task(B b, E e, C c, P&& part = P()) {
 
     // only myself - no need to spawn another graph
     if(W <= 1 || N <= part.chunk_size()) {
-      std::for_each(beg, end, c);
+      task_wrapper([&](){
+        std::for_each(beg, end, c);
+      });
       return;
     }
 
@@ -36,8 +40,29 @@ TF_FORCE_INLINE auto make_for_each_task(B b, E e, C c, P&& part = P()) {
       size_t chunk_size;
       for(size_t w=0, curr_b=0; w<W && curr_b < N; ++w, curr_b += chunk_size) {
         chunk_size = part.adjusted_chunk_size(N, W, w);
-        launch_loop(W, w, rt, [=, &c, &part] () mutable {
-          part.loop(N, W, curr_b, chunk_size,
+        launch_loop(W, w, rt, [=, &c, &part,&task_wrapper] () mutable {
+          task_wrapper([&]{
+            part.loop(N, W, curr_b, chunk_size,
+              [&, prev_e=size_t{0}](size_t part_b, size_t part_e) mutable {
+                std::advance(beg, part_b - prev_e);
+                for(size_t x = part_b; x<part_e; x++) {
+                  c(*beg++);
+                }
+                prev_e = part_e;
+              }
+            ); 
+          });
+        });
+      }
+
+      rt.corun_all();
+    }
+    // dynamic partitioner
+    else {
+      std::atomic<size_t> next(0);
+      launch_loop(N, W, rt, next, part, [=, &c, &next, &part,&task_wrapper] () mutable {
+        task_wrapper([&]() {
+          part.loop(N, W, next, 
             [&, prev_e=size_t{0}](size_t part_b, size_t part_e) mutable {
               std::advance(beg, part_b - prev_e);
               for(size_t x = part_b; x<part_e; x++) {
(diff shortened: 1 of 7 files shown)


taskflow/algorithm/for_each.hpp
```diff
@@ -36,16 +38,18 @@ TF_FORCE_INLINE auto make_for_each_task(B beg, E end, C c, P&& part = P()) {
...
-              std::advance(beg, curr_b - prev_e);
-              for(size_t x = curr_b; x<curr_e; x++) {
-                c(*beg++);
+        launch_loop(W, w, rt, [=, &c, &part,&task_wrapper] () mutable {
+          task_wrapper([&]{
+            part.loop(N, W, curr_b, chunk_size,
+              [&, prev_e=size_t{0}](size_t curr_b, size_t curr_e) mutable {
+                std::advance(beg, curr_b - prev_e);
+                for(size_t x = curr_b; x<curr_e; x++) {
+                  c(*beg++);
```
olologin, line 47: Shouldn't we execute wrapper for this thread?

taskflow/algorithm/for_each.hpp
```diff
@@ -54,32 +58,34 @@ TF_FORCE_INLINE auto make_for_each_task(B beg, E end, C c, P&& part = P()) {
...
-            std::advance(beg, curr_b - prev_e);
-            for(size_t x = curr_b; x<curr_e; x++) {
-              c(*beg++);
+      launch_loop(N, W, rt, next, part, [=, &c, &next, &part,&task_wrapper] () mutable {
+        task_wrapper([&]() {
+          part.loop(N, W, next, 
+            [&, prev_e=size_t{0}](size_t curr_b, size_t curr_e) mutable {
+              std::advance(beg, curr_b - prev_e);
+              for(size_t x = curr_b; x<curr_e; x++) {
+                c(*beg++);
```
olologin, line 67: Shouldn't we execute wrapper here?

taskflow/algorithm/for_each.hpp
olologin: I mean for general case, nobody knows if user-supplied wrapper can be skipped for parent thread.

olologin:
I resolved merge conflicts and rebased it on the current master. Also fixed problems that I highlighted before.

@tsung-wei-huang could this be reviewed and merged? We are using this patch internally and it helps us with performance for situations when tasks have to be executed in some environment with specific FPU settings, etc.

tsung-wei-huang:
@olologin I am reviewing it today and will get it merged if no problem :)

olologin:
Hmm, I will look into these two failing tests.

olologin:
Tests should be fixed now. But TSAN job failure seems like something that is not caused by our change. Correct me if I am wrong.

tsung-wei-huang:
I am thinking to combine TaskWrapper with Partitioner, which allows more flexible extension in the future using Partitioner. For example:

```
tf::Executor executor(tc);
std::exception_ptr x;
tf::StaticPartitioner part(1, [&](auto &&task){
  try {
  task();
  }catch(...){
    x = std::current_exception();
  }
});

std::atomic<int> count = 0;
tf::Taskflow taskflow;
taskflow.for_each_index(
    0, 100, 1, [&](int i)
    { count++; },
    tf::StaticPartitioner(1)
);
if(x)
   std::rethrow_exception(exceptionPtr);
```

What do you think?

bradphelan:
I'm not sure that the coupling is correct. For example there are algorithms that you provide that do not take a partitioner such a parallel scan but still require a task wrapper. 

image

tsung-wei-huang:
It depends on the algorithm itself. parallel-scan does not support partitioner due to its algorithm nature.

bradphelan:
> It depends on the algorithm itself. parallel-scan does not support partitioner due to its algorithm nature.

That is my point. Not all algorithms use partitioners but you are proposing integrating a feature with partitioners which would make the feature unavailable to algorithms that do not use partitioners.

tsung-wei-huang:
@bradphelan @olologin thank you for contributing to this great feature. I have merged it - as I said, don't worry about the c++20 version of the failed test. I will fix it later.

PR #507, opened 2023-09-13
Test for issue #506

bradphelan:
This shows issue #506 failing

### DIFF
--- a/unittests/test_semaphores.cpp
+++ b/unittests/test_semaphores.cpp
@@ -9,55 +9,55 @@
 
 void critical_section(size_t W) {
 
-  tf::Taskflow taskflow;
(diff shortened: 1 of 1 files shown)


tsung-wei-huang:
I am looking into that. thanks!

tsung-wei-huang:
@bradphelan The problem happens because `tf::Semaphore` or `tf::CriticalSection` is not thread-safe. In your case, multiple subflows will try to add their tasks into the critical section. Does this make sense?

CodeHarbour-Genie (commented):
Review for unittests/test_semaphores.cpp:
Additions: 122
Deletions: 78
AI Genie analysis: <generator object Client._stream at 0x000002007453AF00>

Issue #266, opened 2020-12-30
Exception handling


robinchrist:
Yeah... once again: Exception handling.

There are a couple of issues about this, e.g. #72 or #193

My proposal:
Make exception handling optional and allow the user to specify the behaviour via (combinable) flags.

Exception handling also shouldn't be limited to a single exception, but rather store all exceptions that were thrown. `future.get()` should return the first exception that was thrown.

I think these two flags would be appropriate to start with:
`TF_ENABLE_EXCEPTION_HANDLING` -> Exceptions are caught and stored in order that `std::abort` does not get called.
`TF_CANCEL_ON_EXCEPTION` -> Only valid in combination with `TF_ENABLE_EXCEPTION_HANDLING`, automatically calls `cancel` once an exception has been caught.

Things that need to be worked out
- Where / how would be the flags be specified?
- How are detached Subflows handled?
- How are nested Subflows handled?
- What exactly gets cancelled? IMHO it would make sense to cancel the whole graph, as it would never be able to complete. However there are also detached Subflows...

What do you think?

tsung-wei-huang:
Exception is supported in the current main/dev branch and will be part of the next 3.7.0 release.

PR #521, opened 2023-10-09
add a way to just let an unhandled exception be unhandled

fran6co:
Allow to disable exception handling, this makes it easier to generate stack traces as it lets the exceptions be unhandled and terminate where they were generated

Related to this https://github.com/taskflow/taskflow/issues/266


tsung-wei-huang:
Thank you for the pull!

PR #532, opened 2023-11-23
add compiler explorer link with simple.cpp on README

weilewei:
This PR adds a compiler explorer link on README with simple.cpp example. This can 
- easily onboard more users without any local environments
- enables both developers and users to perform live demo and quick verification
- reduces communication overhead among different local setups
- and more...


tsung-wei-huang:
Thank you very much!

PR #536, opened 2023-12-22
Revised the interface of Closure Wrapper for Partitioner

tsung-wei-huang:
Previously, the pull request #505 extended a task wrapper from the algorithm method itself. This pull request integrate the task wrapper interface into the partitioner object to improve the extensibility directly on the partitioner class itself. 

In #505, we need: 

```
tf::Executor executor(tc);
std::exception_ptr x;
/// Capture the floating point unit configuration
FPUSaver saver;
auto task_wrapper = [&](auto &&closure)
{
  try {
    /// Apply the floating point unit configuration ( and undo on destruction
    FPUApplier applier(saver);
    closure();
  }catch(...){
    x = std::current_exception();
  }
};
std::atomic<int> count = 0;
tf::Taskflow taskflow;
taskflow.for_each_index(
    0, 100, 1, [&](int i)
    { count++; },
    tf::StaticPartitioner(1), task_wrapper);
if(x)
   std::rethrow_exception(exceptionPtr);
```

In this pull request, we no longer need a separate task wrapper but integrate it into the partitioner object:

```
tf::Executor executor(tc);
std::exception_ptr x;
/// Capture the floating point unit configuration
FPUSaver saver;
std::atomic<int> count = 0;
tf::Taskflow taskflow;
taskflow.for_each_index(0, 100, 1, 
    [&](int i) { count++; },
    tf::StaticPartitioner(1,  [&](auto && closure)
    {
        try {
        // Apply the floating point unit configuration ( and undo on destruction
        FPUApplier applier(saver);
        closure();
      } catch(...){
        x = std::current_exception();
      }
    }
);

if(x)
   std::rethrow_exception(exceptionPtr);
```

This will allow future extension on the algorithm to directly work on the partitioner instead of creating many other template parameters to the method (e.g., `tf::Taskflow::for_each(beg, end, [](auto){}, partitioner, wrapper1, wapper2, wrapper3, ...)`).

@bradphelan @olologin - could you please have a look and see if you can accommodate such change to your existing codebase? The modification should be straightforward and will allow future extension in a more concise manner.


bradphelan:
I think this is a problem because we don't want users at moduleworks
modifying our exception capture code or FPU management via our wrappers but
we do want them to be able to choose the partitioner for individual
algorithms. This means we will need to create wrapper factories for all the
partitioners.

Somehow I feel this would be better served as an independent parameter or
as a property of the executor that only requires setting once.

olologin:
@bradphelan 
Making it a property of an executor will prevent people from running child tasks in the parent environment. What if in some component we will need to apply specific FP settings? For example boost::voronoi requires correct work of FP infinities, which we normally prohibit for most of our code.

bradphelan:
Maybe then the tf::Runtime object ?

https://taskflow.github.io/taskflow/classtf_1_1Runtime.html

You can have different *Runtimes* associated with the same scheduler.

tsung-wei-huang:
@bradphelan @olologin 

> I think this is a problem because we don't want users at moduleworks modifying our exception capture code or FPU management via our wrappers but we do want them to be able to choose the partitioner for individual algorithms. This means we will need to create wrapper factories for all the partitioners. Somehow I feel this would be better served as an independent parameter or as a property of the executor that only requires setting once.

I know it can take some modification but the change should be very tiny and it is the right way to keep good API in the long run. Combining the task wrapper with partitioner has many advantages: (1) No need of another template argument in the algorithm method. What if I need another partition-specific argument `A` in the future? Introducing another argument will require user to specify partition size, task wrapper, until that `A` (2) in the future, we can easily extend the partitioner interface to take different closures. (3) everything is unified.

olologin:
@tsung-wei-huang Sorry for late response. It works, we have switched our code to partition-based wrapper. Thank you!

tsung-wei-huang:
@olologin thank you for letting me know it works!

PR #548, opened 2024-02-05
C++-20 compilation error fix

doocman:
C++ 20 has removed the std::allocator<void>-specialisation, quick wokaround is to replace the type alias with the intended type. Fixes issue 539.


tsung-wei-huang:
@doocman thank you for the pull request to fix the issue!

PR #549, opened 2024-02-05
guard from calling eventDestroy on NULL cuda events

doocman:


tsung-wei-huang:
@doocman thank you for the help!

Issue #566, opened 2024-03-27
Specify g++ version, or disable compile without certain compiler support


wilderfield:
**Please describe your feature request.**

This code base uses std::inclusive_scan which does not exist in g++-7.

Is there a way we can document the supported compilers, or maybe better perform some check in CMake that checks for the minimum required compiler support?

tsung-wei-huang:
The Readme does mention the G++ version needs to be [at least v8](https://github.com/taskflow/taskflow?tab=readme-ov-file#supported-compilers)

PR #569, opened 2024-04-02
Update CMakeLists.txt

wilderfield:
Set default CXX Standard to C++17 as it is the minimum required.

User can override it for experimentation.

Update the minimum required compiler support for GNU/GCC and AppleClang per documentation.

Addresses: https://github.com/taskflow/taskflow/issues/566


tsung-wei-huang:
Thank you for the catch!!!

PR #579, opened 2024-04-19
Show the correct function to be called: corun_all() instead of join().

### LINKED ISSUE #577: Doc issue: Documentation shows wrong function to call.
The examples at https://taskflow.github.io/taskflow/AsyncTasking.html#LaunchAsynchronousTasksFromARuntime , showing how to use `tf::Runtime::async()` then go on to suggest calling `rt.join()`. I think this is a copy-paste issue from above, where the `join()` function was called on sub-flows. But `tf::Runtime` does not have a `join()` member function. I *think* the code snippets should show calling `corun_all()` instead -- which would then also be consistent with the documentation that intersperses the code snippets.

bangerth:
I think this should fix #577.


tsung-wei-huang:
Thank you!

Issue #578, opened 2024-04-19
buildtest failed on armel: "undefined reference to `__atomic_store_8'"


juliangilbey:
**Describe the bug**
The package (version 3.6.0) fails to successfully compile the testsuite on the armel architecture.

The build proceeds without problem (though lots of warning messages about parameter passing changing in GCC 7.1) until:

```
[  7%] Linking CXX executable test_work_stealing
cd /home/jdg/taskflow-3.6.0+ds/buildtest/unittests && /usr/bin/cmake -E cmake_link_script CMakeFiles/test_work_stealing.dir/link.txt --verbose=1
/usr/bin/c++ -g -O2 -ffile-prefix-map=/home/jdg/taskflow-3.6.0+ds=. -fstack-protector-strong -fstack-clash-protection -Wformat -Werror=format-security -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -D_TIME_BITS=64 -Wdate-time -D_FORTIFY_SOURCE=2 -O3 -DNDEBUG -Wl,-z,relro -Wl,-z,now CMakeFiles/test_work_stealing.dir/test_work_stealing.cpp.o -o test_work_stealing 
/usr/bin/ld: CMakeFiles/test_work_stealing.dir/test_work_stealing.cpp.o: in function `std::__atomic_base<long long>::store(long long, std::memory_order)':
/usr/include/c++/13/bits/atomic_base.h:481:(.text+0x347c): undefined reference to `__atomic_store_8'
/usr/bin/ld: /usr/include/c++/13/bits/atomic_base.h:481:(.text+0x3490): undefined reference to `__atomic_store_8'
/usr/bin/ld: /usr/include/c++/13/bits/atomic_base.h:481:(.text+0x34f0): undefined reference to `__atomic_store_8'
/usr/bin/ld: /usr/include/c++/13/bits/atomic_base.h:481:(.text+0x3504): undefined reference to `__atomic_store_8'
/usr/bin/ld: /usr/include/c++/13/bits/atomic_base.h:481:(.text+0x3564): undefined reference to `__atomic_store_8'
/usr/bin/ld: CMakeFiles/test_work_stealing.dir/test_work_stealing.cpp.o:/usr/include/c++/13/bits/atomic_base.h:481: more undefined references to `__atomic_store_8' follow
[...]

PR #580, opened 2024-04-21
Handle the need to explicitly link to libatomic on some archs (#578)

juliangilbey:
This patch appears to fix the undefined reference to `__atomic_*_8` issue (#578).  I might have specified `${ATOMIC_LIBRARY}` more than needed, but everything seems to compile with this patch (at least when applied to v3.6.0 rather than HEAD) both on an amd64 machine and on an armel machine.


tsung-wei-huang:
Thank you for the help!

PR #585, opened 2024-05-07
operator new() will never return a 'nullptr'.

bangerth:
`operator new` will throw `std::bad_alloc` if it cannot allocate enough space. It is not necessary to check the returned value against `nullptr`.


tsung-wei-huang:
Thank you, @bangerth !

PR #609, opened 2024-08-08
Make `for_each_index` actually heterogeneous

blonded04:
**Problem description**: https://github.com/taskflow/taskflow/issues/610
Function `for_each_index` has following signature: `template <typename B, typename E, typename S, typename C, typename P> Task FlowBuilder::for_each_index(B beg, E end, S inc, C c, P part);` which implies, that range might be heterogeneous: for example, `beg` and `end` can be of type `size_t`, and `inc` can be `int`.

However, it's not the case, because C++ compilers fail to deduce type on functions `is_range_invalid` and `distance`, which are used under the hood of `for_each_index`, because they are homogeneous. So, before this pull-request, something like following example wouldn't compile, because compiler fails to deduce types of arguments for homogeneous functions `distance` and `is_range_invalid`:
```
template<typename F>
void parallel_for(size_t from, size_t to, F&& f) {
    tf.for_each_index(from, to, 1, std::forward<F>(f));
}
```

**Result**
This PR fixes this problem and adds a unit-test with an example of heterogeneous call to `for_each_index`

### DIFF
--- a/taskflow/utility/iterator.hpp
+++ b/taskflow/utility/iterator.hpp
@@ -5,17 +5,21 @@
 
 namespace tf {
 
-template <typename T>
-constexpr std::enable_if_t<std::is_integral<std::decay_t<T>>::value, bool>
-is_range_invalid(T beg, T end, T step) {
+template <typename B, typename E, typename S>
+constexpr std::enable_if_t<std::is_integral<std::decay_t<B>>::value && 
+                           std::is_integral<std::decay_t<E>>::value && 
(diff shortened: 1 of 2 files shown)


blonded04:
Now unit-tests pass! They didn't pass before because I added wrong unittest that thinks that parallel algorithm `for_each_index(l, r)` is over `[l; r]`, while it's actually over `[l; r)`

tsung-wei-huang:
@blonded04 Thank you for your help! This is a great fix!!!

PR #618, opened 2024-09-05
feat(examples): Add names to tasks in simple

HRXWEB:
Add names to tasks A, B, C, and D for better identification in Graphviz

Add taskflow dump for profile.


tsung-wei-huang:
@HRXWEB Thanks. This is a good idea.

PR #623, opened 2024-09-11
fix(docs): fix wrong varibale name and executable name in doxygen dox

HRXWEB:
# benchmark_taskflow.dox

As shown in the content of [benchmark/CMakeLists.txt](https://github.com/taskflow/taskflow/blob/master/benchmarks/CMakeLists.txt), all benchmark executable files now have the bench_ prefix. However, the documentation was not updated accordingly. I have made some updates to reflect these changes.

# executor.dox

same variable name is wrong.

# Additionally

The contents of docs are generated by doxygen and m.css is tracked in repo, so need to update also. But the environment of mine is different from the default. Please update it by yourself @tsung-wei-huang

### DIFF
--- a/doxygen/cookbook/executor.dox
+++ b/doxygen/cookbook/executor.dox
@@ -98,7 +98,7 @@ tf::Executor executor;  // create an executor
   // ... 
 
   // run the taskflow
-  executor.run(f);
+  executor.run(taskflow);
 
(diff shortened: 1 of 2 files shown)


tsung-wei-huang:
Thank you for this fix!!!

PR #624, opened 2024-09-14
bugfix(doxygen): update conditional tasking in doxygen/cookbook/conditional_tasking.dox

HRXWEB:
The original example code contains the following issues:

1. The expression `++c < 10 ? 0 : -1` should return 1 in the false case; otherwise, the executor will not be able to find the successor task to execute, and the final result of count will be 10.
2. A source task should be added; otherwise, the taskflow will encounter a "no source task" error.
3. Task A should also be a `loop` task and a predecessor task for (A, B).

Additionally, I have added some code to help newcomers debug and visualize the graph more easily.

Lastly, could someone please explain the purpose of the if(i) code block:
https://github.com/taskflow/taskflow/blob/master/doxygen/cookbook/conditional_tasking.dox#L515-518

 Thanks advance.


tsung-wei-huang:
@HRXWEB Thank you for the catch! 👍

PR #605, opened 2024-07-07
Doxygen fix for Executor class

tjhei:
The current doxygen documentation does not show the class ``Executor``, see https://taskflow.github.io/taskflow/annotated.html. I don't know for sure which line of code inside the executor.hpp file confuses doxygen, but excluding the implementations from being parsed by doxygen fixes the issue. For this, I had to let doxygen define a DOXYGEN macro.

PR #625, opened 2024-09-16
Fix the `Executor` class not appearing in generator docs, patch warnings

mosra:
Original issue on the m.css repo: mosra/m.css#250

This PR excludes the `Executor` implementation from Doxygen processing as proposed in #605, just with a more unique name (@tjhei I listed you as a co-author). Additionally fixes warnings related to filesystem hierarchy by enabling FULL_PATH_NAMES + STRIP_FROM_PATH. That ultimately results in the same output as before, but with an effect that the directory structure doesn't have to be guessed from incomplete information by m.css. The warnings manifest as the following as of mosra/m.css@083c7a87dcca80027a4bc9c5e6db3bab4babffb0:

> WARNING:root:potential issue: the parent of algorithm/ is taskflow/ which is not a prefix, you may want to enable FULL_PATH_NAMES together with STRIP_FROM_PATH and STRIP_FROM_INC_PATH to preserve filesystem hierarchy

Finally, adds workarounds for certain Doxygen parsing warts, where `@param` is put together with the immediately following `@code` into a single paragraph, causing warnings like below. The paragraph ends up being empty and is thus removed, again resulting in the same generated output as before, just without the warnings.

> WARNING:root:classtf_1_1Runtime.xml: inline code has multiple lines, fallback to a code block

(Just for completeness -- inline code higlighting [can look for example like this](https://doc.magnum.graphics/magnum/types.html), note the C++ and GLSL keywords being bold. [Documentation on how to do that is here.](https://mcss.mosra.cz/documentation/doxygen/#code-highlighting))

### DIFF
--- a/doxygen/Doxyfile
+++ b/doxygen/Doxyfile
@@ -172,7 +172,7 @@ INLINE_INHERITED_MEMB  = NO
 # shortest path that makes the file name unique will be used
 # The default value is: YES.
 
-FULL_PATH_NAMES        = NO
+FULL_PATH_NAMES        = YES
 
 # The STRIP_FROM_PATH tag can be used to strip a user-defined part of the path.
 # Stripping is only done if one of the specified strings matches the left-hand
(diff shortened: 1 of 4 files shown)


tsung-wei-huang:
Thank you, @mosra  and @tjhei ! The `#def` is not perfect but I understand it's a problem of doxygen. Feel free to create pull requests if you find a better solution in the future. For now, let's stick with this workaround.

PR #626, opened 2024-09-22
use ADL to support custom distance()/advance() in for_each

tjhei:
If there is no customized distance() or advance() function visible, this will still call the functions in ``std::``, otherwise the users' functions are called.

### DIFF
--- a/taskflow/algorithm/for_each.hpp
+++ b/taskflow/algorithm/for_each.hpp
@@ -11,6 +11,10 @@ auto make_for_each_task(B b, E e, C c, P part = P()) {
   using B_t = std::decay_t<unwrap_ref_decay_t<B>>;
   using E_t = std::decay_t<unwrap_ref_decay_t<E>>;
 
+  // use ADL to allow for custom distance() and advance() functions:
+  using std::distance;
+  using std::advance;
+
   return [=] (Runtime& rt) mutable {
 
     // fetch the stateful values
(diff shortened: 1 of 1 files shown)


tjhei:
I hope that makes sense. If you approve, I am happy to do this for other parts of the code (transform, reduce, etc.).

tsung-wei-huang:
@tjhei thank you for the pull request. I think the best way to do this is let application customize their iterators that support their own `++` methods. Current interface templates the iterator type and should already provide a solution (IMHO) better than overwriting `std::advance` with custom `advance`.

bangerth:
@tsung-wei-huang Are you suggesting that user codes overload `std::distance()` and `std::advance()` for their own iterator classes? I don't think the standard allows that -- you can't just put things into `namespace std` with the exception of a small number of specific names.

tjhei:
> Are you suggesting that user codes overload `std::distance()` and `std::advance()` for their own iterator classes?

If you use ADL as in the patch I propose, one can write their own ``distance`` (and ``advance``, but the latter is less critical) that lives outside the namespace std, see https://stackoverflow.com/a/47977571

@tsung-wei-huang proposes to implement all iterator traits/functionality that ``std::distance`` requires. I think, that for an O(1) ``std::distance`` call, the iterator would need to fulfill everything stated here: https://en.cppreference.com/w/cpp/named_req/RandomAccessIterator including all the type traits and functionality it is derived from.

My suggestion seems to be much simpler and comes without any drawback in the library (the patch is quite simple).

Issue #631, opened 2024-09-30
misuse in unit-test of utility


NanaGlutamate:
https://github.com/taskflow/taskflow/blob/3b512a3172281412f8ea6c0eee3086b644277197/unittests/test_utility.cpp#L243

NanaGlutamate:
seems should be `REQUIRE(size == std::unique(uuids.begin(), uuids.end()) - uuids.begin())` or somethings like this

tsung-wei-huang:
Thank you very much!

PR #632, opened 2024-09-30
fix use of std::unique in test_utility.cpp

NanaGlutamate:


NanaGlutamate:
fix #631

tsung-wei-huang:
Thank you very much!

PR #656, opened 2025-01-14
Improve freelist item distribution with a simple hash

andatr:
### Improve freelist item distribution across queues

The `class Freelist` previously used the `Node*` address to distribute items between queues. However, this approach did not account for the fact that `Node` objects are aligned to 8 bytes. As a result, some queues received more items than others, leading to uneven distribution. For example, with concurrency set to 8, only 1 out of 8 queues is populated. With concurrency set to 32, only 4 queues are populated.

To address this, I explored several solutions, including dividing the address by 8. Ultimately, the implemented solution, which uses a simple hash calculation, provided the best results in terms of performance.

This issue becomes noticeable only with a large number of tasks, so I created a specialized example to test and analyze the impact of the fix on performance. The example deliberately avoids using the Partitioner, as it obscures the issue by grouping tasks and reducing the total number of method calls.

On my PC with 32 logical cores (16 physical cores with hyper-threading), the example demonstrated a performance improvement from an average of 994,303 microseconds to 733,910 microseconds, resulting in a 24% improvement. While this is a significant gain, the Partitioner reduces the relevance of this fix for typical workloads, as it limits the number of calls to this method.

Notably, the total runtime of the test suite increased by 0.5 seconds on my system. This is likely because most tests do not involve a large enough number of tasks to expose the original issue, making the fix less impactful in such scenarios.

### Example

```
#include <taskflow/taskflow.hpp>
#include <taskflow/algorithm/for_each.hpp>

void for_each(size_t N) {

  tf::Executor executor;
  tf::Taskflow taskflow;

  struct Vec3
  {
    float a;
    float b;
    float r;
  };

  std::random_device rd;
  std::mt19937 gen(rd());
  std::uniform_real_distribution<float> dist(0.0f, 1.0f);
  std::vector<Vec3> values(N);
  for (auto& v : values) {
    v.a = dist(gen);
    v.b = dist(gen);
    v.r = 0.0f;
  }

  for (size_t i = 0; i < N; ++i) {
    taskflow.emplace([&values, i]() {
      values[i].r = values[i].a + values[i].b;
    });
  }

  auto beg = std::chrono::high_resolution_clock::now();
  executor.run(taskflow).get();
  auto end = std::chrono::high_resolution_clock::now();
  std::cout << std::chrono::duration_cast<std::chrono::microseconds>(end - beg).count() << std::endl;

  // Simulate output of useful data to prevent unnecessary optimizations
  float result = 0.0f;
  for (auto& v : values) {
    result += v.r;
  }  
  std::cout << result << std::endl;
}

int main(int argc, char* argv[]) {

  for_each(10000000);
  return 0;
}
```


tsung-wei-huang:
Thank you very much for the pull request. Yes, pointer (`Node*`) is aligned to 8 bytes and that can cause the freelist to have serious contention when the number of free tasks is inserted into freelist. I slightly refactored your code and rebase it to the newest dev/master branch.

andatr:
Happy to help!

PR #664, opened 2025-02-14
Enable use of 'Taskflow::Taskflow' for all consumption styles

afuller-TT:
Create an ALIAS target that matches the exported target name (TaskFlow::Taskflow).  This allows consumers to be agnostic of how Taskflow was brought into the build.  ie: a consuming project may use `Taskflow::Taskflow` and it will work with an external packaging of Taskflow, and also when built-on-the-fly (eg: via git submodule).


tsung-wei-huang:
Thank you for the contribution!

PR #666, opened 2025-02-20
Use 'requires' keyword to support C++20 features

yumeno-yan:
This pull request updates the code to use C++20's `requires` keyword in place of SFINAE-based constraints.  This change brings several benefits:

1.  **Improved Readability**: `requires` offers a cleaner and more intuitive way to express template constraints.  In contrast, SFINAE requires constraints to be embedded within the template parameter list using `enable_if_t` and a special `nullptr`.  By separating the constraints from the template parameters, `requires` improves code clarity and makes the intention of the template easier to understand.

2.  **Enhanced Type Safety**: Using `requires` allows the compiler to more easily validate template parameters against the specified conditions, providing better type-checking and more precise error messages during compilation. When a constraint fails, the compiler can now show exactly which concept check failed, making it easier to pinpoint and fix issues. This helps catch potential errors early, reducing the likelihood of subtle bugs that might arise from more complex SFINAE conditions.

3.  **Aligning with C++20 Standards**: By using `requires`, the code is more in line with modern C++ standards.  This approach embraces C++20 features, ensuring that the code remains up-to-date and takes advantage of the latest language improvements.

### DIFF
--- a/taskflow/core/async.hpp
+++ b/taskflow/core/async.hpp
@@ -78,9 +78,14 @@ inline void Executor::_tear_down_async(Worker& worker, Node* node, Node*& cache)
 // ----------------------------------------------------------------------------
 
 // Function: silent_dependent_async
+#if __cplusplus >= TF_CPP20
+template <typename F, typename... Tasks>
+requires all_same_v<AsyncTask, std::decay_t<Tasks>...>
+#else
 template <typename F, typename... Tasks,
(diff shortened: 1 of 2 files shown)


tsung-wei-huang:
@yumeno-yan thank you for this pull request - and sorry for the delay. This is an amazing change as it simplifies a lot of meta programming. However, I will keep this open now as doing this will confuse the doxygen when generating correct template tags. We plan to migrate everything to C++20 in the near future (v4.0)

PR #671, opened 2025-03-18
Fix MSVC macro conflict

N3VIN:
This errors occur because the Windows` min `and `max` macros conflict with `std::min` and `std::max`.


tsung-wei-huang:
Thank you for the fix!

PR #676, opened 2025-04-09
Issue #506: task_isolation - rebase feature on the new branch, add unit tests, add Lazy wrapper

olologin:
Solves #506 
Of course I just show my changes, let me know if you see some area for improvement.
Would be great to upstream this eventually.

### DIFF
--- a/taskflow/algorithm/lazy.hpp
+++ b/taskflow/algorithm/lazy.hpp
@@ -1,0 +1,168 @@
+#pragma once
+
+#include "../taskflow.hpp"
+#include <atomic>
+
+#if __cplusplus >= TF_CPP20
+#include <condition_variable>
+#include <mutex>
+#endif
+
+namespace tf
+{
+// Lazy initializer that can be used with taskflow without deadlocking
+template <class T>
+class Lazy
+{
+	/// The internal implementation that all instances share
+	struct LazyImpl
+	{
+		/// The cached result
+		// we cannot actually use std::variant here, since we cannot construct a function return
+		// value in-place that way
+		alignas(std::max(alignof(T), alignof(std::exception_ptr))) unsigned char m_data[std::max(
+			sizeof(T), sizeof(std::exception_ptr))];
+		enum class DataType : unsigned char
+		{
+			NONE = 0,
+			CALCULATING = 1,
+			VALUE = 2,
+			EXCEPTION = 3
+		};
+		/// Thread safety handling
+		std::atomic<DataType> m_dataType;
+		/// The function used to generate the result
+		const std::function<T()> m_fn;
+		tf::Executor& m_ex;
+
+#if __cplusplus < TF_CPP20
+		std::condition_variable m_cv;
+		std::mutex m_mtx;
+#endif
+
+		LazyImpl(std::function<T()> f, tf::Executor& ex) : m_dataType(DataType::NONE), m_fn(f), m_ex(ex) {}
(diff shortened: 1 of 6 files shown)


bradphelan:
I like this very much. I would be tempted to clean the lazy class by extracting a ValueOrException class. The inling of the placement new and other related machinery next to the taskflow machinery makes understanding either more complex. They can be seperated. I have two before and after examples.

Before:  https://godbolt.org/z/qsEbf569q

After: https://godbolt.org/z/dr175s35q

The extraction is a simple class which can be tested in isolation. Perhaps the name of the class should be TryCatch because that is really what it is modelling.

```cpp
// New helper class that encapsulates value or exception handling
template <typename T>
class ValueOrException {
  alignas(std::max(alignof(T), alignof(std::exception_ptr)))
  unsigned char m_data[std::max(sizeof(T), sizeof(std::exception_ptr))];
  enum class State { NONE, VALUE, EXCEPTION } m_state = State::NONE;

public:
  ~ValueOrException() {
    reset();
  }

  /// Emplaces the result of calling fn directly into
  /// the store. If the function fails then the exception
  /// is stored.
  void emplace_fn(std::function<T()> fn) {
    try {
      new (m_data) T(fn());
      m_state = State::VALUE;
    } catch (...) {
      new (m_data) std::exception_ptr(std::current_exception());
      m_state = State::EXCEPTION;
    }
  }

  T* get() {
    if (m_state == State::EXCEPTION) {
      std::rethrow_exception(*std::launder(reinterpret_cast<std::exception_ptr*>(m_data)));
    }
    return std::launder(reinterpret_cast<T*>(m_data));
  }

  bool has_value() const {
    return m_state == State::VALUE;
  }

  bool ready() const {
    return m_state != State::NONE;
  }

  void reset() {
    if (m_state == State::VALUE) {
      std::launder(reinterpret_cast<T*>(m_data))->~T();
    } else if (m_state == State::EXCEPTION) {
      std::launder(reinterpret_cast<std::exception_ptr*>(m_data))->~exception_ptr();
    }
    m_state = State::NONE;
  }
};
```

olologin:
Not sure about the implementation, but regarding API:
I think it would be better to get rid of Executor::isolate method and allow user to use TaskArenaApplierRAII directly (maybe with better name).

This way user can implement some kind of cooperative mutex locking that does not block the calling thread and instead runs corun_until until mutex is available.
Basically it would be an alternative like

```
taskflow::Executor ex;
...
std::mutex m;
taskflow::scoped_lock l(m, ex); // If m.lock() succeeds - enter new task arena for the current thread, if not - ex.corun_until until m.lock() succeeds.

// Code that requires single-threaded execution
...
```

tsung-wei-huang:
@olologin , I am looking into this pull - and thank you again for this great contribution. A few suggestions:

1. Can you rebase it to the current master branch that reflects the newest release 3.10? We incorporated quite a few performance-related changes to executor, but I think those changes do not have much conflict with your `TaskArena` proposal.
2. I suggest using [std::enable_shared_from_this](https://en.cppreference.com/w/cpp/memory/enable_shared_from_this) for  `TaskArena`, which makes it more clear to understand the shared ownership of `TaskArena`.
3. I suggest renaming `TaskArenaApplierRAII` to `ScopedTaskArena` to better align with C++ naming convention.
4. I agree with @bradphelan 's suggestion of `ValueOrException`

Feel free to let me know if you have any thoughts.

PR #689, opened 2025-05-07
fix CMakeLists.txt

yumeno-yan:
The current code uses the **__cplusplus** macro to detect C++ versions, but MSVC does not update this macro by default (it remains **199711L**). This causes incorrect version detection in Visual Studio.

This PR adds **/Zc:__cplusplus** to CMake compile options for MSVC, ensuring __cplusplus reflects the actual C++ standard.


tsung-wei-huang:
@yumeno-yan thank you very much!

PR #690, opened 2025-05-10
Toggle try/catch in executor

IkeOTL:
Disable try/catch in executor if TF_DISABLE_EXCEPTION_HANDLING defined


tsung-wei-huang:
Thank you for this pull request! I indeed miss this part of try/catch.

PR #694, opened 2025-05-15
update compiler versions in docs and printouts

m-fila:
Updating the minimal compiler versions in cmake printouts and documentation. This is assuming the checks in cmake are still correct


tsung-wei-huang:
Thank you very much!

PR #727, opened 2025-10-20
Switch this_worker to mechanism that does not depend on DLL boundaries

olologin:
`thread_local` variables are not really "thread" local on windows, they are "dll x thread" local, so if you initialize thread_local variable in one DLL, and then try to read it in another - you will run into problems. Therefore using old mechanism with unordered_set storing ids of all threads is more reliable, because now ids are tied to executor instead of "thread x dll" context.


olologin:
@tsung-wei-huang 
C++ standard has no idea what "dll" or "module" is. So there is no clear definition how statics and thread_local variables should work across modules.
BTW statics on windows are also module-local (each module sees its own version of static when it the same header-only definition is linked into multiple dlls), not global.

Anyway, I believe this should be fixed in taskflow simply because we cannot fix the way windows works with static/thread_local variables. I would probably just have #ifdef that enables this unordered_set-based implementation on all platforms if user wants it. Or user can configure it to enable it only on windows himself.
I would probably even enable it by default in your repository, I did not check performance but I doubt it is that performance-critical :) Because testing all kinds of different combinations of #ifdef is not something good/easy.

tsung-wei-huang:
@olologin thank you for the pull request. I slightly modified your implementation to remove the need of latch.

PR #736, opened 2025-11-21
core/taskflow: Add module names to dot dump if available

### LINKED ISSUE #735: Dump module task name in label instead of "[m1]"
When using module tasks, giving a name to the task is meaningless and instead the module block is labeled just as "[m1]", which is also appended to the task name in the module that composes it.
Instead it should just use the module's name if possible.

Attached an output from `parallel_data_pipeline` example, left is current and right is what I propose.

Image

DaividFrank:
Closes #735


tsung-wei-huang:
Thank you!

Issue #742, opened 2025-12-11
Misleading documentation about cudaGraph and cudaGraphExec lifetimes


adamreeve:
The GPU tasking documentation says:

> the executable graph is an independent entity and has no lifetime dependency on its parent CUDA graph

But I found that if I allowed the `cudaGraph` to be deleted and tried to update the parameters of tasks in a `cudaGraphExcec`, I would get errors from CUDA about invalid values.

For example I modified the `cudaGraph.Update.Memset` test cast to demonstrate:
```c++
TEST_CASE("cudaGraph.Update.MemsetWithFreedGraph") {

  const size_t N = 1024;

  tf::cudaGraphExec exec;
  tf::cudaTask t1;

  auto vec = tf::cuda_malloc_shared<int>(N);
  
  {
    tf::cudaGraph cg;
    t1 = cg.memset(vec, 0x01, N*sizeof(int));

    exec = tf::cudaGraphExec(cg);
  } // cg freed here

  tf::cudaStream stream;

  stream.run(exec).synchronize();
  
  for(size_t i=0; i<N; i++) {
    REQUIRE(vec[i] == 0x01010101);
  }

  exec.memset(t1, vec, 0x0F, N*sizeof(int));

  stream.run(exec).synchronize();
  
  for(size_t i=0; i<N; i++) {
    REQUIRE(vec[i] == 0x0F0F0F0F);
  }

  tf::cuda_free(vec);
}
```

This fails with an error:
```
RROR: test case THREW exception: [cuda_graph_exec.hpp:326] invalid argument (cudaErrorInvalidValue) - failed to update memset parameters on cudaTask [type=Kernel]
```

I assume this is because although the `cudaGraphExec` can be used after the `cudaGraph` is freed, the `cudaTask` is no longer valid, which wasn't obvious to me from the documentation.

If this is expected behaviour, I think a note could be added in the "Attention" section to mention that the parent `cudaGraph` needs to be kept alive in order for its tasks to be updated.

adamreeve:
Fixed in #745

PR #745, opened 2025-12-16
Update GPU tasking documentation

### LINKED ISSUE #741: Incorrect "Update an Executable CUDA Graph" example in GPU tasking documentation
In the two lines where the graph is run, this example uses:
```
stream.run(stream).synchronize();
```
But this should be:
```
stream.run(exec).synchronize();
```

I would have just made a PR to fix this but it looks like this is defined in three different places in the repo and I wasn't sure what the correct process is to change the docs.

adamreeve:
Fixes #741 and #742


tsung-wei-huang:
thank you!

PR #750, opened 2026-01-03
Add C++ modules support

mikomikotaishi:
This PR adds support for C++ modules. This is enabled with `TF_BUILD_MODULES` and creates a module `tf`.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -66,6 +66,13 @@ if(TF_BUILD_SYCL)
   add_subdirectory(sycl)
 endif()
 
+# -----------------------------------------------------------------------------
+# modules examples
+# -----------------------------------------------------------------------------
+if(TF_BUILD_MODULE)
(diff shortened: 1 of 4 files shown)


tsung-wei-huang:
I think we don't need to module dsl which is just a fun project but never used in practical application. Would you mind removing dsl from the module list? Thanks!

mikomikotaishi:
OK, I just included it because it was in the header list so I thought it was relevant to include.

tsung-wei-huang:
@mikomikotaishi thank you for this excellent work. If it is convenient for you, I would appreciate any help with adding one or two examples of using modules. We can place in under a new folder `examples/module` and include a new `examples/module/CMakeLists.txt` to compiler your module example code only when `TF_BUILD_MODULE` is enabled.

mikomikotaishi:
OK, I can try writing a few examples. Do let me know if you would like me to update the README or include additional information elsewhere on how to use the feature.

Issue #755, opened 2026-02-11
Observer triggers (prologue, epilogue) for module tasks


nolankramer:
**Please describe your feature request.**

It's sometimes useful and convenient to capture module performance characteristics (such as scheduling overhead), or trigger callbacks when a module is invoked or finished.

It would be nice if observers also triggered for module tasks. This would allow API callers to do some useful work when aggregate tasks are preparing to run, finishing.

tsung-wei-huang:
@nolankramer  this is a good suggestion. Would you mind creating a pull request for this? Thanks!

PR #758, opened 2026-02-18
Observer triggers (prologue, epilogue) for module tasks #755

nolankramer:
Addresses #755


nolankramer:
@tsung-wei-huang Ready for your review

PR #783, opened 2026-04-13
Feat fill and generate

Hatem-707:
Added fill and generate algorithms and their bounded versions.


Hatem-707:
@tsung-wei-huang is there a reason to use different types for the begin and end iterators. Also I think it would be helpful to add a ticket to track parallel algorithms implemented and perhaps priority for the unimplemented onse

tsung-wei-huang:
The reason is to enable stateful iterator passing so one task can initialize data structure needed while its successor can launch the algorithm. See [capture iterator by reference](https://taskflow.github.io/taskflow/ParallelIterations.html). In reality, we may have something like:

```cpp
taskflow.for_each_index(std::ref(beg), end, 1, [](auto i) {});
```

Hatem-707:
Thanks for the explanation. I edited my implementation to support this and added new tests accordingly

tsung-wei-huang:
Thank you for the contributions!

PR #786, opened 2026-04-22
Add pkg-config support

GaetanLepage:
## Summary

Taskflow currently only provides CMake's `find_package` for discovery. This PR adds a `taskflow.pc` file so that downstream projects using pkg-config (or build systems like Meson that prefer it) can also discover Taskflow.

## Changes

- Add `taskflow.pc.in` template
- Add `configure_file` and `install` rules in `CMakeLists.txt` to generate and install `taskflow.pc` to `${CMAKE_INSTALL_LIBDIR}/pkgconfig`


tsung-wei-huang:
Thanks!

GaetanLepage:
Thanks for the quick reviews! Any plans for a (minor) release in the short/mid term?

PR #798, opened 2026-06-30
No longer test for C++11.

bangerth:
In looking at https://github.com/dealii/dealii/issues/19188, I found a place where the current code base still checks whether the C++ standard is at least C++11. That's no longer necessary because Taskflow requires C++20 anyway.

I also looked at the other part of the `#if` here, testing for whether `defined(UINTPTR_MAX)` exists. From C++11 to C++26, the standard defines `std::uintptr_t` as an optional type. So it's not guaranteed that it always exists, and using `defined(UINTPTR_MAX)` is still necessary.


tsung-wei-huang:
@bangerth thank you for the pull request!

PR #803, opened 2026-07-20
Making CudaGraph-constructor SFINAE with `requires`

### LINKED ISSUE #753: cudaGraph constructor is not SFINAE-friendly
**Describe the bug**
Doing something like:
```C++
if constexpr(std::constructible_from<tf::cudaGraph_t, std::string>) {
  return tf::cudaGraph_t("hello");
else {
  return tf::cudaGrapg_t();
}
```
cause a compilation error as the constructor in tf::cudaGrapg_t just delegates to it's creator. Adding a simple `requires(std::invocable<Creator, Args...>)` to the constructor would probably solve the issue.

fior512:
Fixes #753, with `requires std::is_invocable_r_v<...>` for SFINAE friendly behavior of `CudaGraphBase` inside `taskflow/cuda/cuda_graph.hpp:550`

### Tests
Replicated with:
```cpp
static_assert(!std::constructible_from<cudaGraph, std::string>, "bug: cudaGraph should not be constructible with std::string");
```
with master branch (`cudagraph_i753.cpp` is a local file for test):
```cpp
cudagraph_i753.cpp:10:17: error: static assertion failed: bug: cudaGraph should not be constructible from std::string
   10 |   static_assert(!std::constructible_from<cudaGraph, std::string>, "bug: cudaGraph should not be constructible with std::string");
      |                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
with this PR: fixed.

### Notes
- Similar to PR #666 in standby due to `requires` and doxygen compatibility

### DIFF
--- a/taskflow/cuda/cuda_graph.hpp
+++ b/taskflow/cuda/cuda_graph.hpp
@@ -545,8 +545,10 @@ class cudaGraphBase : public std::unique_ptr<std::remove_pointer_t<cudaGraph_t>,
   Constructs a `cudaGraph` object by passing the given arguments to the executable CUDA graph creator
 
   @param args arguments to pass to the executable CUDA graph creator
+  @note participates in overload resolution only if `Creator` invocable with `ArgsT...` with cudaGraph_t
   */
   template <typename... ArgsT>
+  requires std::is_invocable_r_v<cudaGraph_t, Creator, ArgsT...>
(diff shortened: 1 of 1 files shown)


tsung-wei-huang:
Excellent pull! Thanks!

PR #805, opened 2026-07-27
Fix typo in Speculative Execution example description

adavidzh:


tsung-wei-huang:
Thank you for catching this!

